PR #6, opened 2020-10-07
Added file copy function

PJEstrada:
Links to docs:

https://diffgram.readme.io/reference#file-copy-2
https://diffgram.readme.io/docs/files


swirlingsand (approved):
Great work!

PR #31, opened 2021-04-15
Github actions Migration

PJEstrada:


PJEstrada:
Tests are finally passing. Issue was something related with github actions steps running in background. It seems that we need to put all the dependent services (default, walrus, frontend, dispatcher) on the same command. Otherwhise we get incosistencies on the test results.

Im merging this now to master

PR #34, opened 2021-04-19
Add redirect to set password on first usage.

PJEstrada:
Make users set the password on first usage. Most of the time user will not set their email provider so this is a necessary step on the initial setup.


anthony-sarkis:
Great work Pablo! Thank you very much :)

PR #35, opened 2021-04-19
Add Azure Connector

PJEstrada:
Adding support for uploading files from azure containers, exporting into azure and managing connections from azure.


PJEstrada:
Azure Connector is now working:

- Support for uploading files from azure container into Diffgram
- Support for browsing the azure container
- Support for testing & creating the azure connection
- Support for sending export from diffgram into an azure container.

.editorconfig
```diff
@@ -0,0 +1,31 @@
+[*]
```
anthony-sarkis, line 1: Is this a config file? Do we want this file as part of merge or ?

walrus/methods/connectors/azure_connector.py
```diff
@@ -0,0 +1,410 @@
...
+                # TODO: Decide, do we want to raise an exception? or just do nothing?
+                log = regular_log.default()
+                log['error']['invalid_type'] = 'File must type of: {} {}'.format(str(images_allowed_file_names),
+                                                                                 str(videos_allowed_file_names))
+                log['error']['file_name'] = opts['path']
+                log['opts'] = opts
+                Event.new(
+                    session = session,
+                    member_id = opts['event_data']['request_user'],
+                    kind = 'aws_s3_new_import_warning',
```
anthony-sarkis, line 131: @PJEstrada Should we make this more generic or update it to be azure specific or?

.editorconfig
PJEstrada: Yes I added this to move a bit in the direction of having a standard way of configuring the editor. This file is (in theory) globally supported by most editors and we can set the indent, spacing and tabs in here for everything. So I think it's a good idea to commit it to force code styling in future. 

For now it has basic configurations but we can extend it more in the future.

anthony-sarkis:
Ok cool sounds good!

walrus/methods/connectors/azure_connector.py
PJEstrada: Yes you're right! Thanks for catching that. I've fixed it here: https://github.com/diffgram/diffgram/pull/35/commits/65dc4e53d981078b6d766bb4de16de3ec208d9cd

anthony-sarkis:
Super thank you!

anthony-sarkis (approved):

PR #42, opened 2021-04-21
Fixed issues with hotkeys

PJEstrada:
Seemed that the real problem was not so much with hot keys but with the selected prop of the instances. When we changed frames the watcher would always reset to false the instance select and that would cause some of the hotkeys for related to selection to stop working (like copy and paste). 

This was cause due to not updating the this.previous_current_frame value in the watcher. Now the select is working properly and the hotkeys are following

I also tested the w,a,s,d the space bar, the escape keys, numkeys 1-9 and they all seem to be working fine.

Please let me know if you are aware of any other hotkeys malfunctioning so I can dive deeper into them here.


anthony-sarkis:
Hey, 
I tested with the directory select in media bar and was able to reproduce issue of hotkeys. Inspecting state it showed that the flag we check `$store.state.user.is_typing_or_menu_open` was remaining in `true` even when deselecting.  Watching the mutations I see that it's only firing one event.
menu is not setting state
I committed https://github.com/diffgram/diffgram/commit/d715b857f669d2f085c12efc03cb211b84a17282 

1. This will cover the `blur` event (no change) case.
2. I moved the "did select something" case into the `change` event so it's more clear the relationship of that (vs having it seperate in the `update:return value` -> which made it look as though it was the mirror of the `focus` when it's not.

anthony-sarkis:
I found a few more
https://github.com/diffgram/diffgram/commit/2dfb2685b0fe0e2a01ed558550e445cb9611af5f
Covers
1. Share menu dialog is clicked outside (not close button)
2. Moves the issue dialog creation closer to the actual interface, for menus that don't fill full screen I think this is probably better

userscript select was missing blur
https://github.com/diffgram/diffgram/commit/b6c151a874a9c522ec8460690b14dd9c4d02196f

Please double check ok with
https://github.com/diffgram/diffgram/commit/f354470975aadbc48ca05dcd90c7a3fa6eaab76e
As far as I can see there isn't anything in instance history that needs to disable this,
if there is, then prefer to put it closer to source like in other commits 
https://github.com/diffgram/diffgram/commit/3e559f7ab035ae4333e8e33947dccfcd520c5bd6

anthony-sarkis:
@PJEstrada   I think this is ready for a fresh look when ready

anthony-sarkis (approved):

PJEstrada:
Hey Anthony thanks for catching the onBlur issue :) I checked the code and its looking good. I added https://github.com/diffgram/diffgram/pull/42/commits/2ce3b599f8a0b8cd1c241a8f1ce6fa851e32d596 which just adds the delete hotkey in the global context so it still works if the mouse is not inside the canvas.

PR #44, opened 2021-04-21
Icons now loading in single host context.

PJEstrada:
Still need to test if they are working in multi host context.


PJEstrada:
Tests now passing. The problem was related to a relative path on the webpack conf and not routing static url to the frontend service. Now it should be working fine. Let me know if there are any issues on this, otherwise we can merge this.

anthony-sarkis:
Thanks Pablo! :)

PR #46, opened 2021-04-23
Decouple Media Core

PJEstrada:
The purpose of getting out media core is to allow the file management to be independent of the annotation UI, and make all the annotations UI's to only receive a single file without consciousness of how that file was fetched.

This is a requirement to have before starting to implement the actual text interface: https://github.com/diffgram/diffgram/pull/45


PJEstrada:
Some of the initial refactor things:

- I'm moving `File_list` out of the `annotation_core` component and into `media_core`. The idea is that `annotation_core` should not be responsible of the file list management.
- I'm moving `fetch_project_file_list` and `fetch_single_file`  into media_core

Some questions to discuss:
- `annotation_core` still has some functions related to tasks like `fetch_single_task`. I would like to move this out of `annotation_ccore`, but I don't think it's a good idea to give `media_core` task responsabilitites. What do you think about moving the task fetching to `annotation_full`. That way `annotation_full` will have the responsability to fetch the task, get the file data (using media core methods) and based on that, decide which annoation UI to render (text or image). Another option would be to create a separate component just for task fetching/management purposes but I don't feel like this could be a component since there is no real UI rendered from this.

PJEstrada:
Latest Updates:

- I've created `file_manager_sheet` component. The idea of this component is that we can eventually transform it into a tabbed panel so that we can move between the file manager and the dataset explorer. This allows to have both functionality in the same sheet.
- I've also moved the `media_core` inside of this component and added a button to maximize the sheet. The idea is that if we want more screen real estate to view file previews and browse to the data, you can press the maximize button and get the full screen version of the file manager. If we click a file, it will minimize and we'll get the usual annotion core view.

anthony-sarkis:
Sounds good

PJEstrada:
I think I have this branch in a stable state now. I've removed media core completly from annotation core and added the `file_manager_sheet` component. I've added a tabbed panel so we can add the explorer feature on the same place without losing the screen real state. For now it just has 1 tab but it can grow to multiple tabs (for example the comparison/explore feature)

image


Also the `file_manager_sheet` no also has the capability of maximizing, so if we want to focus 100% on file management/analysis we can just use the entire screen to do so. When you click a file it automatically minimizes so we can go straight into labeling or inspecting the file.

image

anthony-sarkis:
Thanks Pablo this is really great!! I think this is a huge step forward for
the new dataset management features too

Let's talk about it a bit on our next call.
Yes I will look at circle CI

PJEstrada:
Cool! Thank a lot Anthony:)

anthony-sarkis:
1) Moved message here so it shows up in full screen too (and just better place for it)
no files match part of media core
<br>

2) Adjusted buttons to try and make better for first time user and use built in style with spacer
https://github.com/diffgram/diffgram/commit/5a691d468bc1635158873301a697ae085e277344
restore down with close
maximize 1

<br>
3) Labels are now computed props in the most parent component.
Not saying it's exactly right yet, but I think it's a step in a cleaner direction for that concern.

image

PR #51, opened 2021-05-05
Add External DB To Install Script

PJEstrada:
Now the install script will ask the user for 2 new options:

1. The Diffgram version tag, by default it will point to the latest version tag.
2. The postgres database url, which will populate `DATABASE_URL` parameter and conditionally change the image tag of postgres to a `scratch` docker image that does nothing (so we have no postgres installed)

### DIFF
--- a/install.py
+++ b/install.py
@@ -3,6 +3,7 @@ import random
 import string
 import base64
 
+
 def create_random_string(length):
     return ''.join(random.choice(string.ascii_lowercase + \
                                  string.digits) for x in range(length))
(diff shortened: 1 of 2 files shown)


install.py
```diff
@@ -138,6 +155,23 @@ def launch_dockers(self):
         print('Diffgram Successfully Launched!')
         print('View the Web UI at: http://localhost:8085')
 
+    def set_diffgram_version(self):
+        version = input('Enter diffgram version: [Or Press Enter to Get The Latest Version]: ')
+        if version == "":
+            self.diffgram_version = 'latest'
+        else:
+            self.diffgram_version = 'version'
```
anthony-sarkis, line 163: We want variable not string here maybe?

anthony-sarkis:
Great work Pablo! :)   
I made one small comment otherwise looks ready to merge
Thank you!

install.py
PJEstrada: You're right thanks for catching that one!

PR #60, opened 2021-05-12
Upload Wizard

PJEstrada:
This new feature aims to provide an easier way to upload files with prelabeled data without requiring any coding at all.

- Creating new components for upload wizard
- Create a new schema mapping concept to map any JSON file to diffgram's format.
- Added stronger bindings to the `input_batch` concept.\
- Added new alembic migrations to support prelabeled data at the batch level.
- Pending to modify some E2E tests as the old upload flow will break with this new code.

Adding this PR as a draft and WIP to give visibility of work progress.


anthony-sarkis:
Awesome work @PJEstrada!

PJEstrada:
Here are the file for testing the UI out: https://www.transfernow.net/dltransfer/?utm_source=202105188YvRtz7z&utm_medium=&utm_content=en

Main features added until now:

1. Completed components for UI Wizard Flow for uploading files 
2. Upload from local with no instances
2. Upload from local Image And video with prelabeled data on csv or json
3. Uploading files without prelabeled data via connections

Still WIP:
- Uploading prelabeled data from connections.
- Updating files from the upload wizard.
- Refactoring the upload dialog on the task template creation
- Docs and examples.
- Saving the model_run and run_id.

PJEstrada:
I've finished both flows on the wizard now. You should be able to test out both the new files and the update files.

- Both support CSV and JSON files
- You should be able to use connections now too.

I'm going to add the model run and model now and then work on some e2e tests and I think we'll be ready to launch this one :)

PJEstrada:
I've added the sheet component and restructured the layout. This is the new look:

image


The idea is to remove scroll by putting the table side to side with the dropzone and add a scollbar at the table level and not at the dialog level so the continue button is always visible. Let me know what you think! :)

anthony-sarkis:
It's a good idea! 

I was thinking too, what if the Dataset selector component is it's own step? Maybe is easier on new users? 
That way the screenshot shown screen can just be focused on the data selection part

"Choose existing dataset"          "Create new dataset"
as the two options? picturing like two buttons centered horizontally and vertically?

Picturing if user clicks "NEW", then maybe it just centers the "name" field in the center and that's only item on page?
Just brainstorming :)

PJEstrada:
> It's a good idea!
> 
> I was thinking too, what if the Dataset selector component is it's own step? Maybe is easier on new users?
> That way the screenshot shown screen can just be focused on the data selection part
> 
> "Choose existing dataset" "Create new dataset"
> as the two options? picturing like two buttons centered horizontally and vertically?
> 
> Picturing if user clicks "NEW", then maybe it just centers the "name" field in the center and that's only item on page?
> Just brainstorming :)

Thanks Anthony! Agree with separating the dataset step! It's usually something people miss out since it's so small currently. I'll try to add something better based on those ideas! :D

anthony-sarkis:
Super! Thanks Pablo! :)

PJEstrada:
Tests are now passing and ready to go. 

Here's the link for some docs related to the feature: https://diffgram.readme.io/docs/importing-files-instances-with-the-upload-wizard

It's mostly a small walkthrough in the UI. Let me know if there's anything else we can add on the docs here. I will work on one small change and will merge later today :)

PR #63, opened 2021-05-12
Improve UI, especially in video and task context - WIP

anthony-sarkis:
Key features

1. Label selector default is now integrated in toolbar. In this context a user can only select one label at a time.
Prior, the rationale for keeping the "open" style was that it was easier to select / see all the labels.
This allows removal of the right panel (all right panel items shifted to left panel).
2. The toolbar now has a revised priority, with the most commonly used buttons in the left most area.
And added overflow. Quantitively there are now 7 buttons moved to new overflow menu. Vertical dividers added to help visually group related functions. Visually looks a lot cleaner.
3. Video tool bar is similarly cleaned up (new overflow menu) and more context is added for frame and duration.
4. Technically, a new toolbar component has been created. All of the toolbar code ~(700 lines) has been migrated to there from annotation core. While this is far from an ideal abstraction, it's a solid step towards cleaner code and smoother feature flags. For example if a user wished to hide the "new instance type" selector there's now a more clean path.
5. In Task mode, the top menu bar is hidden by default, and additional buttons are added for navigation.
For new annotation users, and especially casual annotation users, this present a much more focused UI with less cluter.
For pro annotators this gives you another 50px of screen real estate for controls and maximizing viewing size.

The overall effect is much more welcome for new users, while maintaining the functionality. 

# New
new video layout

# Old
image

### more menu
image
image


anthony-sarkis:
New look for media core:

image


image

image

Filters
<br>

image

Full screen context (From Pablo's work!)
<br>
image

PJEstrada:
This looks awesome!! 🎉🎉💪🏼🥳

anthony-sarkis:
> This looks awesome!! 🎉🎉💪🏼🥳

Thank you!!! :) I'm excited for this release!!

PJEstrada:
Hey Anthony! First of all this is great UI improvements! The annotation_core looks much cleaner now and gives the platform a whole new feel.

I was doing a bit of review/e2e tests and here are some of the things I caught and fixed based on the doc.

1. Not sure if it's a priority, but when drawing instances, and then adding a line of code to the annotation core. The hot reload removes the drawn instances, also the image dissapears
image
You can replicate it by drawing 3 instances and then adding a console.log to annotation core.
2.The undo/redo was not updating the canvas. I fixed that one here: https://github.com/diffgram/diffgram/pull/63/commits/4ce9f844c4f6461849d8d1a5b5463e32d737de8e . If there are any other cases where the canvas is not updating please let me know, this was the only one I managed to be able to reproduce.
3. Maybe it is a good idea to reset panel sizes on file change? I made the pannel a bit  bigger for a video, and then changed file. The next file was a bit bigger to the UI ended up lookin like this:
image
4. Defer issues are fixed here: https://github.com/diffgram/diffgram/pull/63/commits/c2bcad28f69e73f4d1fc92145f86f3e7bd518b64
5. This one is just a UI thing, and mainly opinion. Maybe we can make the toolbar and the main topbar to be full width? Another idea is to make the main background white too, I think it can achieve the same effect. Let me know what you think of this one
image

I'm just pending to fix those hot reload issues where the file and instances are lost. Otherwise I think this is a great improvement of the UI thanks a lot Anthony!

PJEstrada:
I managed to figure out the hot reload issue. It seems to get fixed when adding this lines to the created() function in annotation_core:
      ```
// Initial File Set
      if(this.$props.file){
        this.on_change_current_file();
      }
      else if(this.$props.task){
        this.on_change_current_task();
      }
```

I know you removed them on your commits and there was a reason for it, maybe we can discuss if we should re add them or find another alternative?

PR #65, opened 2021-05-15
Job page clean up

anthony-sarkis:


anthony-sarkis:
New Job page is cleaner
image

PR #67, opened 2021-05-18
👻 Ghost Objects - Video Now 10x easier for new annotators 

anthony-sarkis:
New "Ghost" instance proposals make it 10x easy to annotate for both casual and pro users.
Now, a user doesn't really need to understand anything about sequences to use sequences.

## Overview
As a user I can draw the the objects.
The system presents context that I can choose to accept or ignore.
There is no need to expressly "decline" proposals, all proposals are declined by default.

## Features
New "Ghost" proposals. Ghost proposals are generated on frame change. Crucially
1) It's based on the last keyframe. A user can go at any pace, from frame by frame to jumping thousands of frames.
2) It's clearly different from normal instances. By using a seperate list and rendering mechanism, this is isolated from the normal instance list. This presents a better "middle ground" between our prior state of not showing _any_ context, and other implementations that just show the full instance and it's confusing what's generated and what's user made.

Feature flag: `label_settings.show_ghost_instances = true`

## Video
https://www.loom.com/share/bc71bd9cab4f4e729f4189ca60595c19

## Visual
ghost objects explained

## Pause Feature
image

## User prompt upon first time feature is available
image

## Future
1) I'd like to extend this idea to user scripts and other things that generate proposals.
2) "Pause" has a lot of room for improvement, eg a "Resume" function
3)  ghost_instance_list.vue is meant to be refactored into a concrete implementation that uses abstractions currently duplicated from instance_list. For the sake of timing I'm ok to do that as a seperate branch if everyone is ok with that too.

## Known issues - None appear super critical for v1 usage

1) Ghost logic is in annotation core but should be a class outside of annotation core
2) Some occasional timing issues with saving, initial keyframe. Any bugs in keyframe list break ghost objects.
3) "Selection" needs to be reviewed in this new context - sometimes the "selected" instance is off
4) Undo/Redo/Resize sometimes behavior unexpectedly in this new context
5) An instance paused prior to being the last key frame in a series isn't well handled

@PJEstrada  please let me know your thoughts :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 
 
-# Diffgram - Supervised Learning Data Platform
+# Diffgram - Open AI Data Platform
 
 ![](./github_assets/overview_diffgram_high_level.PNG)
 
@@ -49,7 +49,8 @@ python install.py
 # After install:  View the Web UI at: http://localhost:8085
 ```
 ### Other Useful Links For Starting Out
-- [Install Guide Compute Engine](https://medium.com/diffgram/tutorial-install-diffgram-in-google-compute-engine-134aae7d8a9b)
+- [Google GCP Install Guide Compute Engine](https://medium.com/diffgram/tutorial-install-diffgram-in-google-compute-engine-134aae7d8a9b)
+- [Azure AKS Kubernetes Install Guide](https://medium.com/diffgram/tutorial-installing-diffgram-on-azure-aks-b9447685e271)
 - [Updating Existing Installation](https://diffgram.readme.io/docs/updating-an-existing-installation)
 - [Development Install Docs](https://diffgram.readme.io/docs/quickstart-installation-of-diffgram-open-core)
 - [Production Install Docs](https://diffgram.readme.io/docs/open-installation-production)
@@ -94,7 +95,7 @@ Security issues: Do not create a public issue. Email security@diffgram.com with 
 * [Streaming](https://diffgram.com/streaming)
 * [Security and Privacy](https://diffgram.com/secure)
 * [Speed Up with AI Userscripts](https://diffgram.readme.io/docs/userscript-examples)
-* Open Core (This Repo!)
+* Open Source (This Repo!)
 * [Integrations](#integrations)
 
 # Speed Ups & AI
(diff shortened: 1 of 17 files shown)


PJEstrada (commented):
This is a great new feature Anthony! It's a very simple speedup idea that will be really useful for users with sequence intensive videos :) Here are some of my comments, questions.

1. It's super smart to just create a copy of the instance on the next frame, most of the time then following an object it can not move so in most cases it will work fine. +1 for integrations with userscript, we could offer mini interpolations for ghost instances which would improve even more the precision.

2. Does this feature affect anything in the image context? I just want to make sure if this is a video specific feature only. If so maybe adding some guards for video mode to prevent things from happening.

3. Maybe it's a good idea to create a reset_ghost_instances() function to use in the context of changing a file or a task.

4. I was also thinking about the attributes here. What strategy are you using for filling the values of attributes? Another idea on this is maybe we can show the attributes with some opacity or add some icon/text to indicate that its not an "offical" instance yet.

5. I got into this weird state when drawing some instances and then accepting a ghost. It seems to have no ID and the changes appear to be saved. I think it can be solved just by triggering the pending changes flag whenever we click a ghost instance maybe?
image

6. Another small thing, I was working in label A. And at frame 3 I got new boxes from the label 'apple'. However when changing to frame 4 there where not ghost instances for this 2 boxes. The 2 boxes where already existing on the system I did not draw them on my usage session.
image
image

7. For the issue creation it seems that I can select ghost instances too. Maybe we can block that from happening on the UI? Also when "selecting" the ghost instance, it seems to get created. I think this is expected but maybe not in the issue reation context? This is more just my opinion but we can discuss it a bit further :)

8. Other from that I think this is a great idea Anthony! It looks really good and saves a lot of time, specially valuable for users in the video context with lots of sequences!

PJEstrada:
I was thinking a bit more on this feature and I think adding a hotkey to toggle on/off the ghost could be a good idea. That way the user could just think something like "Let's see what Diffgram suggests... (presses ghost hotkey)" and they can get a quick overview of the ghost suggestions and turn it off once they're done with them

anthony-sarkis:
Thanks Pablo! :) I'm working on the items listed now I'll link to a doc when more complete

anthony-sarkis:
Hey Pablo, thanks for all the great catches and ideas!

I made this doc with the details: 
https://docs.google.com/document/d/1ehe8i7Iv0RxsjmcetEZRUFcdTWnhZ_A9jAZb17r0XEQ/edit
(Anyone in diffgram can edit doc, but external is view only for now)

### A quick summary of updates made:
Reset fix : https://github.com/diffgram/diffgram/commit/2c06c0fbd059ad479f16b82d3ad3d206399f250b
Save fix: https://github.com/diffgram/diffgram/commit/4abf7303a67162aab21a8ec8d21ea80950ee6f1d 
Issue creation context: https://github.com/diffgram/diffgram/commit/bb2f2c52cd13a29de9a81901e3973c87fd15263f

### And of outstanding TODOs from it:
[ ] TODO show the attributes with some opacity or add some icon/text to indicate that its not an "official" instance yet.
I think I’m ok with leaving this for V2 if you are too

Yes already exists -> “G” key
[ ] TODO What’s missing I think is some kind of UI visual to let a user know if they are on or off.
V2 I think if ok with that too?

PJEstrada (approved):

anthony-sarkis:
End to end tests passed on last significant commit.

anthony-sarkis:
Thank you @PJEstrada  for approving! And good discussion on call

PR #70, opened 2021-05-24
bugfix: change dummy image to tianon/true

PJEstrada:
On some docker versions scratch is a reserved word, giving errors to some users when pulling the images.


anthony-sarkis:
Thanks Pablo! :)

PR #74, opened 2021-05-25
tiny:  fix test - context now that we default it to minimized

anthony-sarkis:


anthony-sarkis:
Cypress testing had a good catch here.

In the context of a newly created script, because there were no external sources, the play button would be disabled until the user clicked or sources or reloaded script.
This tiny fix should do it:
https://github.com/diffgram/diffgram/pull/74/commits/46227d306ccdd63d4e0b9dd43ea8d8b957c7486a#diff-bfad5bdf601f0ab8a5210fa0b4973199ceabfada2b1bf40efe4b8c1914b2e5b8R424

#### Also, re-ordered error messages so they are above fold. 
(Fix after design change with UI pushed this component lower down)

reordered stuff
syntax error example

PR #89, opened 2021-06-03
fix: extension extraction validation over None types

PJEstrada:
When the original file name cannot be determined the walrus should not crash but just ignore type extraction.


anthony-sarkis:
Thanks!

PR #96, opened 2021-06-07
fix: Hotkeys bug with vuex

PJEstrada:
This problem did not occured often, but was able to figure it out the following way:

Since we are persisting the vuex store, there is a chance that we can refresh the screen while `set_user_is_typing_or_menu_open` is `true`. For example, refreshing the screen while creating an issue. This creates an invalid initial state which makes hotkeys stop working because the flag was not resetted. 

Now this is fixed by resetting `set_user_is_typing_or_menu_open` on the mounted function of annotation core.


anthony-sarkis:
Great catch Pablo!! :) :) :)

PR #107, opened 2021-06-15
Query Engine V1

PJEstrada:
This version will support basic non-nested expressions with and, or and comparator operators. Based on that it will create a SQL query that can be excuted with SQLAlchemy and return the file list objects to the user.

Features for this version:
- Add support for filtering of files by label count
- Adds support for filtering files by metadata value
- Add support to the parser to interpret correctly and/or staements
- Created the initial grammar definitions for a LALR parser.
- Created the `QueryCreator`, `QueryExecutor` and `DiffgramQuery` classes. 
- Created the `SqlAlchemyQueryExecutor` class to build sql alchemy queries from a string respecting the grammar.

### DIFF
--- a/default/methods/annotation/labels/labels.py
+++ b/default/methods/annotation/labels/labels.py
@@ -128,52 +128,7 @@ def labelRefresh(project_string_id):
         project = Project.get_project(session, project_string_id)
         directory = project.directory_default
 
-        # TODO use shared file get thing...
-
-        # TODO use project based permissions
-        working_dir_sub_query = session.query(WorkingDirFileLink).filter(
-            WorkingDirFileLink.working_dir_id == directory.id,
-            WorkingDirFileLink.type == "label").subquery('working_dir_sub_query')
-
-        # Caution, don't do "state != "removed" here,
-        # Since we may have removed label files
-        # With active instances, and still need this for colour map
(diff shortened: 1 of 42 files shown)


PJEstrada:
Hey @anthony-sarkis !

The query engine is almost ready for merge. I'm going to work on some e2e tests and I think we should be good to go.

I wanted to task if you could review the code and specially the files:

https://github.com/diffgram/diffgram/blob/cc8763f841474a13ff1e09370fb2e9bb1773d011/shared/query_engine/sqlalchemy_query_exectutor.py
https://github.com/diffgram/diffgram/blob/cc8763f841474a13ff1e09370fb2e9bb1773d011/shared/query_engine/query_creator.py
https://github.com/diffgram/diffgram/blob/cc8763f841474a13ff1e09370fb2e9bb1773d011/shared/query_engine/sqlalchemy_query_exectutor.py
https://github.com/diffgram/diffgram/blob/cc8763f841474a13ff1e09370fb2e9bb1773d011/default/methods/source_control/file/file_browser.py

I feel this are the ones that have most of the querying logic and could add attack vectors if we some permissions validations. So if you could see it with more "security" eyes that would be great so we can add any additional validations to prevent unauthorized users from sending queries.

PR #109, opened 2021-06-17
fix: when no label templates available there was an unhandled exception.

PJEstrada:
current labels is undefined if not label templates have been created on the project. I also added a link to the label creation section when clicking the empty label list selector.


anthony-sarkis:
Good going Pablo! 😊

PR #121, opened 2021-06-24
Public project support

PJEstrada:
This new feature allows projects to be public and seen by users with no diffgram account.

The approach to accomplish this is to extend a bit the logic in router.js and add an additions metadata key to the routes called `available_on_public_project`. All routes with `available_on_public_project === true` should be allowed to be rendered if the project provided in the URL is public.

Please let me know if you find any issues with this mechanism for permissions or if you have any additional comments on it.


PJEstrada:
Public projects are now working!

For now you should only be able to access the Studio of the project along with the file manager and dataset explorer:

image


I think this can be enough for people to browse through the files. We can disscuss if there are any other parts of the system that we can show to the public for these types of projects.

PR #132, opened 2021-07-01
Create Missing Labels on Upload Wizard

PJEstrada:
When the wizard does not find the labels, add a button to create all the missing ones so the user can continue with the flow.

This helps specially for big uploads where you would have to manually create 100s or 1000s of labels one by one. This button removes the need to create them manually.

### DIFF
--- a/default/methods/source_control/file/file_browser.py
+++ b/default/methods/source_control/file/file_browser.py
@@ -543,7 +543,7 @@ class File_Browser():
         # outside of limits...
 
         if self.metadata['file_view_mode'] is None or \
-            self.metadata['file_view_mode'] not in ["changes", "annotation", "home", "task", "explorer"]:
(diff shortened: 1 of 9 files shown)


PJEstrada:
I added several optimizations on the Upload Wizard to support large files:

- Usage of `Object.freeze()` for the large instance list uploaded. Current benchmark is that it support more than 850,000 instances on a single JSON file. There 3-6 seconds wait time on some stages (like verifying that all the files exists). But this can vary depending on how many files there are in the JSON file. For the current test we had 102,000+ files in the same JSON file.
- Used memoization on some processing steps to avoid the usage of `filter`, since that greatly added time complexity to the payload creation.
- Removed the `Async.all()` since it can get the machine out of resources when the file count is to high. Even though is a bit slower to send one by one, we avoid too much resource consumption on the client, and possible error due to rate limiting.
image

PR #136, opened 2021-07-06
Dataset Explorer Page number key name

PJEstrada:
On the recent changes to the pagination, the page number key was changed. There were some pending places that needed the name change on the dataset explorer which was causing the pagination not to work properly


anthony-sarkis:
Great!

PR #140, opened 2021-07-08
Optimizing Github Runners

PJEstrada:
Testing different sizes and hosted/non hosted combinations


PJEstrada:
Reduced runner to optimize compute costs of testing suites.

PR #142, opened 2021-07-12
Upload progress was not updating properly

PJEstrada:
UI issue made counter not update correctly on large videos. Now we keep a separate counter per file upload ID and add them up on a computed prop


anthony-sarkis:
Thanks Pablo!!!

PR #166, opened 2021-07-20
Connection Testing Improvements

PJEstrada:
Add better connection testing both on the install.py script as well as the connection testing on the UI.

This will add permissions testing for S3, GCP and Azure providers.


PJEstrada:
Now Install script will show the following message for connection testing: 

image

PJEstrada:
Now the install script will show the following messages:

image

PJEstrada:
For connections now we can see more detailed errors on permission when doing the test connection button:

image

anthony-sarkis:
> Now the install script will show the following messages:
> 
> image

This is great Pablo!!!!   Now we can catch these connection and credential setup stuff right away instead of it "looking like it's working" but that part potentially being incorrectly configured!! Super work!

PR #167, opened 2021-07-21
Improve Loading States File Wizard

PJEstrada:
This will help users know when validations are in progress and avoid confusion on UI getting stuck or something similar.


PJEstrada:
I have added better loading message on each step of the processs:

image


Now we implement a threaded download of the json payload to the server, while the frontend pings until the download status is completed. This minimizes the risk of timeouts with huge uploads and improves the overall resilience of the upload wizard.

Will wait for all tests to pass and merge afterwards

PR #168, opened 2021-07-22
Improve Error Displays

PJEstrada:
Now we can catch Network errors and 500 errors


PJEstrada:
Network errors are now stored in vuex and based on that we display a global snackbar to alert the user of server disconnection:

image

PJEstrada:
Now we also have the setting: `DIFFGRAM_ERROR_SEND_TRACES_IN_RESPONSE`

Which will add traces to responses so UI can display them:

image

PR #170, opened 2021-07-23
System Events v1

PJEstrada:
Based on new needs for better logging external system events like:

- New Installations
- Versions Tracking
- Installations Fingerprints
- Cloud Providers
- Startup and Shutdown Time


PJEstrada:
Branch should now be ready to merge. Main features added:

- Added fingerprint, os and version as environment variable settings.
- Added a SystemStartupChecker class that we can extend for more validations
- Inside the system startup added the validations for fingerprint, version, and os existing. System will not startup unless these values are set to prevent malfunctioning of the system events and ensure better monitoring across installations.
- All the above environment settings are setup using the install.py script (or can be set manually via the ENV variable)

Please let me know if there are any questions or concerns with this one. I'll wait for tests to pass and merge afterwards

PR #178, opened 2021-07-27
Fix batch splitted upload in azure

PJEstrada:
Fixes batched upload failing when using azure backend


anthony-sarkis:
Thanks Pablo!!

PR #186, opened 2021-07-29
Fix View Only Mode State Across Projects

PJEstrada:


PJEstrada:
The view mode getting stucked is now fixed.

Main problems was the Vuex store persisting the public_project module due to a small bug. This vuex module should not be persisted now so that the public_project attribute gets updated on each refresh.

PR #192, opened 2021-07-30
Upload Wizard: support for updates via filename

PJEstrada:
Add the ability to read JSON files and fetch the diffgram file using the original_filename.


anthony-sarkis:
Super!! Thank you very much Pablo!

PR #193, opened 2021-08-04
Small bug fix - upgrade to latest call back

anthony-sarkis:
append-icon-cb was deprecated and causing this function to not work
did manual test and appears to be working as expected

context is that now that we default to requiring password, this was preventing a user from copying and pasting or inspecting their password.

this is a small bug fix will merge once tests pass


PJEstrada:
Awesome thanks for catching that one Anthony :)

PR #197, opened 2021-08-05
Removed the alert on retry log input table

PJEstrada:
Quick fix:
Removed the alert on retry log as it can break some styles of the table.

Just a visual thing, the error alert can make people thing an error has happened when it really is just more info on the upload status.


anthony-sarkis:
Cool

PR #198, opened 2021-08-05
Prevent Silent Deletion of Instances in Annotation_Update

PJEstrada:
In some cases the frontend can have unexpected bugs which makes the instance_list to not be sent completely to the backend.

Old versions assume that if instance was not present, it should be deleted, which caused unexpected silent deletion in some cases.

Now we double check that all the instances are available on the request payload before executing any logic in Annotation_Update

### DIFF
--- a/shared/annotation.py
+++ b/shared/annotation.py
@@ -370,13 +370,15 @@ class Annotation_Update():
         return self.new_added_instances
 
     def __check_all_instances_available_in_new_instance_list(self):
-        if not self.init_existing_instances:
+        if not self.do_init_existing_instances:
             return
         new_id_list = []
         for inst in self.instance_list_new:
             if inst.get('id'):
                 new_id_list.append(inst.get('id'))
 
+        print('NEW ID LIST', new_id_list)
+        print('instance_list_existing', self.instance_list_existing)
         ids_not_included = []
         for instance in self.instance_list_existing:
             if instance.id not in new_id_list:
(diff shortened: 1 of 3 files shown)


PJEstrada:
Now when this cases happen we will see a message similar to this:
image


Finishing e2e tests and we should be good to go!

PJEstrada:
E2e tests and unit tests are all ready, waiting for them to pass and for any feedback before merging :)

shared/annotation.py
```diff
@@ -369,6 +369,34 @@ def instance_template_main(self):
...
 
+    def __check_all_instances_available_in_new_instance_list(self):
+        if not self.do_init_existing_instances:
+            return True
+        new_id_list = []
+        for inst in self.instance_list_new:
+            if inst.get('id'):
+                new_id_list.append(inst.get('id'))
+
+        print('NEW ID LIST', new_id_list)
```
anthony-sarkis, line 380: suggest removing print before merging unless we want to log it fully

anthony-sarkis (approved):

shared/annotation.py
```diff
@@ -371,7 +371,7 @@ def instance_template_main(self):
 
     def __check_all_instances_available_in_new_instance_list(self):
         if not self.do_init_existing_instances:
-            return
+            return True
```
anthony-sarkis, line 374: Good catch :)

PJEstrada: Yes Thanks for noticing that one Anthony :)

PR #199, opened 2021-08-06
Silent Deletion Improvements

PJEstrada:
This improvements build on top if 2 realizations:

- If the file cache was manually/maliciously corrupted via a script, direct sql, etc. The user can get in a state where the file cannot be saved.
- If for some reason the file has was not regenerated re-saving the file won't soft delete  it.

So basically adding 2 additions:

1) Add a reload and regenerate cache button, so that if the above case happen at leas there is a way for the user to get back the correctly saved data and get back to annotating.
2) When we find a case of an instance not existing in the cache, we regenerate the hash to prevent the case where the instance does not match the hash, that way when the user refreshes and saves again the instance should have the correct hash.

### DIFF
--- a/default/methods/source_control/file/file_cache_regen.py
+++ b/default/methods/source_control/file/file_cache_regen.py
@@ -1,0 +1,111 @@
+try:
+    from methods.regular.regular_api import *
+    from methods.source_control.file.remove import remove_core as file_remove_core
+    from methods.source_control.file.file_browser import File_Browser
+
+except:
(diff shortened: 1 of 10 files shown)


PJEstrada:
Now when there's a mismatch on the cache and persisted instance list we will see a screen like this one:

image


I've added more descriptive errors and when the user click the button we will trigger a cache regeneration for the current frame or image. This will prevent further loss of data if multiple users are working on the same file at the same time

PR #218, opened 2021-08-11
Add Label Count Report Support

PJEstrada:
Approach is to add a new flag/column to the report template called `group_by_labels`. This will change the query so that the groupby also groups by label_file_id.

### DIFF
--- a/default/methods/report/report_runner.py
+++ b/default/methods/report/report_runner.py
@@ -601,14 +601,14 @@ class Report_Runner():
         else:
             self.log['error']['base_class_string'] = "base_class_string is None"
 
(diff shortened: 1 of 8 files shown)


PJEstrada:
Added the bar chart report grouped by labels and file:

image


This is a pretty detailed report so it can easily grow huge. I limited the max amount of files to 20, so when there are more than 20 files you'll get this alert to download CSV instead.

image

PJEstrada:
Generated Report: 

image

anthony-sarkis:
Amazing!!! Looks great Pablo!!

PR #225, opened 2021-08-12
Fix Duplicate Hashes Cases Annotation Update

PJEstrada:


PJEstrada:
Hey @anthony-sarkis ! I think the validation should be working ok now. I wanted to request a review from you since this part is always delicate to change. Also if you could try to manually break it by doing crazing pasting on the UI that would help a lot too :) I've tried this and up until now I think we are good.

I also added several unit tests and a complete e2e test to catch this error from happening again. 

Let me know if you have any questions or concerns with the code :)

anthony-sarkis:
Thanks, I'll take a look. I know it's been a long day for you Pablo thank
you very much for all your efforts!

PJEstrada:
Thanks Anthony!! 😀

anthony-sarkis:
I tried copy and pasting a bunch and it seemed to handle it as expected!

image

Refreshing the page shows it's only saving one as expected.
For future (low priority because it should be a rare unintentional state) I added this https://github.com/diffgram/diffgram/issues/233

If I have two windows open and I try adding in one, then updating in the other it fires the expected alert.

So in terms of end user QA it appears all well!

shared/annotation.py
```diff
@@ -399,6 +397,17 @@ def __check_all_instances_available_in_new_instance_list(self):
...
         return True
 
+    def append_new_instance_list_hash(self, instance):
+
+        if instance.soft_delete is False:
+            self.new_instance_dict_hash[instance.hash] = instance
+            return True
+        return False
+
+    def order_new_instances_by_date(self):
```
anthony-sarkis, line 407: @PJEstrada  When do we want to call this `order_new_instances_by_date` in annotation.py?

default/tests/shared/test_annotation.py
```diff
@@ -35,6 +36,222 @@ def setUp(self):
...
+        ann_update.instance_list_new = inst_list
+        print('orders', ann_update.instance_list_new)
+        ann_update.order_new_instances_by_date()
+
+        self.assertEqual(ann_update.instance_list_new[0], inst3)
+        self.assertEqual(ann_update.instance_list_new[1], inst2)
+        self.assertEqual(ann_update.instance_list_new[2], inst1)
+
+
+    def test_special__removing_duplicate_instances_in_new_instance_list(self):
```
anthony-sarkis, line 207: @PJEstrada  GREAT!!! :)   🎉
And that's a great used of added counts! glad to see that working in there

anthony-sarkis:
It's warning as expected :) 

image

Great work Pablo! :) 
And the tests look super comprehensive!

shared/annotation.py
PJEstrada: Hey Anthony! Thanks for all the feedback. I've done this commit to add it just before the cache check https://github.com/diffgram/diffgram/pull/225/commits/07ae4a044f6592031eff262dffd3a1736061ea36

default/tests/shared/test_annotation.py
PJEstrada: Thanks Anthony!! :)

PJEstrada:
Super thanks for the review Anthony! I'll wait for the tests to pass and deploy this on Monday :)

PR #239, opened 2021-08-16
fix: avoid race conditions and concurrent updates on annotation update

PJEstrada:
To reduce cache errors occurrences:

1. Added the with_for_update() to instance list
2. Await the save method on frame changes instead of a sleep() time as sometimes the sleep can be less than the time the update takes, specially on slow connections.

### DIFF
--- a/frontend/src/components/annotation/annotation_core.vue
+++ b/frontend/src/components/annotation/annotation_core.vue
@@ -6270,6 +6270,8 @@ export default Vue.extend( {
         edges: edges,
         version: undefined,
         root_id: undefined,
+        previous_id: undefined,
+        next_id: undefined,
         attribute_groups: {...instance_to_copy.attribute_groups}
       };
 
@@ -6363,6 +6365,7 @@ export default Vue.extend( {
         }
 
       }
+
 
       const current_frontend_instances = instance_list.map(id => id);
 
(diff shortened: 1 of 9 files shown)


frontend/src/components/video/video.vue
```diff
@@ -693,11 +696,7 @@ export default Vue.extend( {
        *    but the channel between video and annotation core is not great
        */
 
-      this.$emit('request_save')
-
-      if (this.has_changed == true) {
-        await this.sleep(200) // give it a chance to save / fire off thing
-      }
+      await this.$props.parent_save()
```
anthony-sarkis, line 699: Thanks goodness!

PJEstrada:
Hey @anthony-sarkis ! I've re requested review, just so you can check the file lock code. I added the defaults for now which is skip_locked = False and nowait = False. That way the query just waits until the file is available with no errors. Let me know what you think about this, and if you feel the nowait should be change to throw an error we can def change it :)

anthony-sarkis:
Thanks Pablo!  Besides item discussed in enterprise thread I think this
looks all good to me!

frontend/src/components/annotation/annotation_core.vue
```diff
@@ -6268,6 +6269,9 @@ export default Vue.extend( {
         nodes: nodes,
         edges: edges,
         version: undefined,
+        root_id: undefined,
```
anthony-sarkis, line 6272: Are there any other things we need to clear? What about creation_ref?

PJEstrada:
Ok so this should be good now :) will wait for the tests and deploy as soon as possible

anthony-sarkis:
Cool! :)

anthony-sarkis:
Thank you!

PR #246, opened 2021-08-19
Add Per Frame Save Loading Flags

PJEstrada:
Before, when users changed frames fast and the save operation took longer than the frame change. They could end up with frames not being saved, this because we only had one global `save_loading` flag per video.

Now we have added the `save_loading_frame` dict to control which frames are saving and not block other frames from saving if a save operation is occuring in another frame

Another bug that in the `add_ids_to_new_instances_and_delete_old` function were were using the reference to this.instance_list instead for the `frame_buffer_dict[request.video_data.current_frame]`. This could cause problems because this.instance_list can change while the annotations are saving. 

Please let me now what you think of these changes and if there's any issue/concern that stands out.

### DIFF
--- a/frontend/src/components/annotation/annotation_core.vue
+++ b/frontend/src/components/annotation/annotation_core.vue
@@ -13,7 +13,7 @@
 
           <toolbar :height="50"
                    :command_manager="command_manager"
-                   :save_loading="save_loading"
+                   :save_loading="this.video_mode ? this.save_loading_frame[this.current_frame] : this.save_loading_image"
                    :annotations_loading="annotations_loading"
                    :loading="loading"
                    :view_only_mode="view_only_mode"
@@ -1089,7 +1089,8 @@ export default Vue.extend( {
       },
 
       annotations_loading: false,
-      save_loading: false,
+      save_loading_image: false,
+      save_loading_frame: {},
       minimize_issues_sidepanel: false,
 
       source_control_menu: false,
@@ -1281,8 +1282,6 @@ export default Vue.extend( {
     }
   },
   computed: {
-
-
     instance_template_dict: function(){
       let result = {};
       for(let i = 0; i < this.instance_template_list.length; i++){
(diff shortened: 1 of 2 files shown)


frontend/src/components/annotation/annotation_core.vue
```diff
@@ -6421,7 +6456,7 @@ export default Vue.extend( {
             current_instance.id = new_instance[0].id;
             current_instance.root_id =  new_instance[0].root_id;
             current_instance.version =  new_instance[0].version;
-            this.instance_list.splice(i, 1, current_instance)
+            instance_list.splice(i, 1, current_instance)
```
anthony-sarkis, line 6459: Good catch!

frontend/src/components/annotation/annotation_core.vue
```diff
@@ -1883,7 +1905,14 @@ export default Vue.extend( {
       this.selected_instance_for_history = undefined;
     },
     warn_user_unload: function(e){
-      if(this.has_changed){
+      let pending_changes_frames = false;
```
anthony-sarkis, line 1908: Great!!

anthony-sarkis:
@PJEstrada  I love it! 

I think this really helps us move away from the assumption of the using having to be on a given frame to save it. 
That's a better user experience.

May be relevant too for shifting interpolation and other functions here in the future. 
This feels like an area that is likely to get future iterations - this is a much better direction.

frontend/src/components/annotation/annotation_core.vue
```diff
@@ -1089,7 +1089,8 @@ export default Vue.extend( {
       },
 
       annotations_loading: false,
-      save_loading: false,
+      save_loading_image: false,
```
anthony-sarkis, line 1092: Is the plan in the future to have concrete flags for text and audio files? If the cases are different enough maybe that's more readable to have it concrete but just double checking

frontend/src/components/annotation/annotation_core.vue
PJEstrada: Thanks Anthony!

frontend/src/components/annotation/annotation_core.vue
PJEstrada: I thin for future file types, I would prefer to move in the direction of having a completely different component built from scratch, since there are still many gotchas and things we don't know yet and too keep bugs and feature changes as isolated as possible. Maybe things that are bery generic, can me moved into a class to be reused by both but in general I would prefer to have separate saves flags and even a completely different componenet for audio/text/lidar for example.

PJEstrada:
Thanks for all the comments Anthony! :) If you're good I can proceed to merge this and deploy

anthony-sarkis:
Oh right right! I was forgetting those new ones will be
separate components. Thanks

anthony-sarkis:
At a high level it all looks good to me!  If you are happy with details etc
please go for it! :)

PJEstrada:
OK perfect thanks Anthony :)

PR #265, opened 2021-08-27
Use Same Frame value and remove soft_delete on hash collisions

PJEstrada:
Use Same Frame value and remove soft_delete on hash collisions

### DIFF
--- a/default/tests/shared/test_annotation.py
+++ b/default/tests/shared/test_annotation.py
@@ -37,6 +37,58 @@ class TestAnnotationUpdate(testing_setup.DiffgramBaseTestCase):
         self.credentials = b64encode("{}:{}".format(self.auth_api.client_id,
                                                     self.auth_api.client_secret).encode()).decode('utf-8')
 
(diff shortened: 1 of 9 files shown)


PJEstrada:
Summary of Changes:
- Removed the part where we delete the old instance id when we find a hash, since in some cases it could end up deleting both instances (the duplicated and the original one)
- Added a guard when we add an instance to the cache_list to prevent from adding the same instance id twice.
- Added a list of duplicated hashes, for testing purposes and for having a bit more information when debugging in the future.
- Added 12+ unit tests and cypress tests to make sure most cases are getting covered
- Added a parallel saving process when pasting instances to next frames, to make sure all pasted instances are saved correctly
- Added some changes to the instance_update() in annotation core to prevent saving the instance inmediately after deleting as this caused some instances to not be saved when deleting multiple instance fast.
- Added the `check_if_pending_created_instance` and the end of `save()` so that we check if any other instances were changed during save process, we reset the flag.

anthony-sarkis:
@PJEstrada Really great work!

PR #267, opened 2021-08-31
Parallel Cypress Tests

PJEstrada:
Improves testing speed by 3x-4x


PJEstrada:
Benchmarks:
- With 2 smaller machine now tests take half the time (16 mins)
- I tested adding up to 7 machine and best time we get is 8 mins. We can eventually add more machine but it will increases costs a bit.
- For now with 2 machines of 4 cores we'll reduce costs a bit more and have a 2x speed gain.

PR #268, opened 2021-08-31
Fetch Frame Buffers In Parallel

PJEstrada:
Add a frame buffer fetch in parallel


PJEstrada:
This code now fetches frames in chunks of 5, and send multiple requests in parallel with a limit of 15 max concurrent requests.

This should improve the speed of the frame buffer as the load now can be distributed across all default instances.

Please let me know if there are any concerns with the code :) otherwise we can merge this

PR #271, opened 2021-09-06
Preview Image Parallel Regen For Sequences

PJEstrada:


PJEstrada:
Hey @anthony-sarkis ! 

The following fixes have beend added:

1. Added the `refresh_missing_preview_images` which checks for any missing previews and hits the API in parallel for recreation. This removed the responsability of generating the images when serializing the sequence which made it slow and caused some row locks on DB, making subsequent updates slower.
2. Added `refresh_sequence_frame_list`, this was due to a bug that new instances that were saved, did not update the frame_list of the sequnece. We had to refresh the page in order to see them.
3. Added the `update_sequence_id_in_cache_list` in annotation.py, this is to add the new sequence_id after the update to the cache. Since the cache list creation happens before the sequence update, the sequence_id was not being saved in the cache list, just in DB. This new function prevents that from happening.
4. Removed the usage of `current_sequence_annotation_core_prop`, for creating new sequences, since this assumed that only 1 sequence was created per save, which is not always the case. This was moved to use  the `add_new_sequence_to_list` which adds all the new sequences returned by annotation_update.

Overall this should make the sequence functions more consistent and performant.  I know there are some comments you added on lines 6764-6766 of annotation core so I wanted to ask if you could double check this code and see if there are any assumptions that are now broken or any cases that have not been taken into account.

Otherwise I think we can merge this and return the normal sequence functionality to annotation update.

anthony-sarkis:
Hey Pablo!

## Good going!! 😊
This is really great work! I tested the sequence preview thing and that is working great. 

Having previews able to reproduce on demand  - and not blocking saving - is a much better direction!

I noticed a couple of things doing manual testing


### Colors
The colors seem to be changing oddly, here in edit mode the colors are fine, 


but in draw they lose their colors


### Sometimes while clicking and drawing the colors flashed oddly too:
not sure if it's related


(Colors appear to be regression vs live version)

### Keyframes gone on file change
As a user, I navigate to the next file, then come back, it seems to appears to lose some of the keyframes.


Before
keyframes before

After going to next file and returning ->
<br>
missing keyframes


<br>
<br>

Those were only 2 things that popped out.

Regarding the code itself, maybe would be good to step through code real quick on screen share please. I think there's one or two parts I'm just not 100% sure on yet and may be faster to knock off any little curiosities that way

PJEstrada:
Hey @anthony-sarkis ! Thanks for your catching those issues!

For the colors problem, did you use copy paste? or any other feature? I'm having trouble replicating it so wanted to know if you have the steps to replicate it or if you have a video maybe showing it?

Same with the sequences, I'm trying to draw some instances on multiple frames, save them and then change file (with the next file button on the toolbar) and in my case I still get the instances. Maybe the label changed and that changed the sequences attached to it? If you have any extra info too on how to replicate it, it would help me a lot :)

anthony-sarkis:
In staging DB http://127.0.0.1:8085/studio/annotate/zenithface?file=38&frame=0&

Video with comments: https://www.loom.com/share/03f28f4421b74e238c2c9dca7b60dfc0

PJEstrada:
Super thanks Anthony I will look further into it, thanks for the video!

anthony-sarkis:
-> Positive direction in that we are now assuming **multiple** sequences saved at once
-> Keyframes are added by frontend pro-actively instead of trying to be replaced by backend
-> backend process still adds the keyframes -> becuase on fresh file load we need to know them among other reasons.  Another benefit is we no longer need to send the whole sequence on response.
-> Default to color map from project or task map (not instance) - whole seperate issue made for that https://github.com/diffgram/diffgram/issues/282

Future iteration thing - I really want the user to be able to create multiple new sequences without waiting for save to return.

PR #279, opened 2021-09-09
Login improvements - small

anthony-sarkis:
Context of debugging -> user not being able to use magic login.
I was unable to reproduce the error - even with an "empty" new user. However I did notice it was possible to get into odd UI states. That could be confusing.

For example if, as a user, I click the back button after doing the magic redeem, it would show this which was confusing.
state of clicking back after magic login is a bit odd it shows login screen

1. Added a guard to prevent running auth if user is already logged in
2. Show state of already logged in

already logged in state

I believe solution is ok. 
An alterative is a navigation guard to prevent entry to this page. Because we use `beforeEnter` , at a quick analysis it seemed like it would have required bigger scope of changes to capture the `back` context. https://stackoverflow.com/questions/50087164/vue-router-how-to-prevent-the-user-back-to-the-login-screen-pressing-the-back-b 
<br>

3. Took opportunity to add loading screen.
While in many cases login is instant, sometimes on first load etc it's possible for this screen to show for a few seconds so this looks better then showing a login screen - and also helps avoid user trying to hit login button again etc. Also avoids more work having to re-work disable flags to remain until next page loads etc.
image

### DIFF
--- a/default/methods/user/account/auth_code.py
+++ b/default/methods/user/account/auth_code.py
@@ -40,10 +40,11 @@ def new(session,
 	# TODO not clear on use of this type of filer, ie if using a code to send to multiple
 	# orgs...
 
-	existing_code = session.query(Signup_code).filter( Signup_code.email_sent_to == email_sent_to, 
-													   Signup_code.type == auth_code_type,
-													   Signup_code.is_available != False,
-														).first()
+	existing_code = session.query(Signup_code).filter( 
+        Signup_code.email_sent_to == email_sent_to, 
+		Signup_code.type == auth_code_type,
+		Signup_code.is_available != False,
+		).first()
 	if existing_code:
 		if existing_code.type == "magic_login":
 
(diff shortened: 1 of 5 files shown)


anthony-sarkis:
@PJEstrada 
For the tests here 
https://dashboard.cypress.io/projects/djxpvm/runs/1930/overview/5ffec487-cf13-49a2-9ba0-610242850209/video
image

it seems like it's already logged in, but we are checking a value there

<br>

I made a suggestion here, please let me know what you think
https://github.com/diffgram/diffgram/pull/279/commits/f099e400bcad729c47568716050cba640b684bec

best
Anthony

PJEstrada (approved):
Hey @anthony-sarkis! Thanks for solving this! I saw the Cypress tests and the components changes and they all seem good to me! If you're ok I think we can merge and close this one :)

anthony-sarkis:
Great, thanks Pablo!

PR #283, opened 2021-09-09
Readme small - Docs only changes

anthony-sarkis:


anthony-sarkis:
docs only change - going to merge :)

PR #285, opened 2021-09-13
Polygon Figures Merge

PJEstrada:
Allows support for attaching multiple figures to the same polygon instance.

This will be accomplished by adding the `figure_id` key to the `points` key in the polygon instance type.


PJEstrada:
Hey @anthony-sarkis ! I think this branch is now ready for review :) 

Main concept added across UI is the concept of `figure_id`:
- Whenever we merge a polygon with another one we add a `figure_id` to each point to know to which polygon it belongs.
- When we unmerge we remove that figure id from the points and if the polygon just has one figure_id left we remove all figure ids to keep it as a "normal" 1 figure polygons.
- So only in the case were we have multiple figures we will have a figure_id in the points list. In all other cases it will not have that key.

Main changes were to support existing feature with this new concept as things like autobordering, moving points and dragging had some gotchas in the case where there are multiple figures.

Let me know if you have any questions or suggestions for the code, otherwise we can merge this one

anthony-sarkis:
I'm looking now

frontend/src/components/context_menu/context_menu.vue
```diff
@@ -496,6 +537,56 @@
           </v-list-item-title>
         </v-list-item-content>
       </v-list-item>
+      <v-list-item
+        link
+        dense
+        data-cy="delete_instance"
```
anthony-sarkis, line 543: may wish to rename these data-cy elements on merge and unmerge buttons

anthony-sarkis:
Amazing work Pablo!!! :)

frontend/src/components/context_menu/context_menu.vue
PJEstrada: Thanks @anthony-sarkis ! I just pushed the fix for this :)

anthony-sarkis:
Cool! :)

PR #289, opened 2021-09-15
Default file level options - WIP

anthony-sarkis:
https://diffgram.readme.io/docs/global-attributes

Prior had to declare a tag. This was not very end user friendly.
Now can declare an attribute group as global and then save it.

This is a WIP there's a few known items to clean up , still need test cases, and a short write up on trade offs, questions etc.

image

image


PJEstrada:
This is great Anthony! Setting the attribute directly is a much straightforward way of doing it and making the system handle the instance creation inner workings.

I only have the question of how  will this look like in the export format? And how will this work for videos? Are the attributes global for all frames or per frame?

PJEstrada:
Main Changes on this PR

- Added the Global Instance Concept (This instance is still inside_the instance list so its managed as an instance)
- There's only one global instance per file on the frontend now.
- Updated the wizard to ask if attribute is global or not.
- Fixed some layout issues, and moved the global attributes list to a separate component for better isolation and separation of conerns.

anthony-sarkis:
Great work Pablo! :)

PJEstrada (approved):

PR #294, opened 2021-09-16
Improve Performance of Report Queries

PJEstrada:
Adding some indexes to key columns that remove sequential scans on certain reports.


anthony-sarkis:
Great work Pablo!!

anthony-sarkis (approved):

PR #302, opened 2021-09-22
Upgrade export task context and general export fixes

anthony-sarkis:
Summary

1. **Make export from tasks work as expected.** Meaning it pulls from tasks not files. This is becuase now that we have streaming, the concept of a file being associated with a singular job is ill-defined. Further it opens the ability to use task orientated filters!
2. It also happens to fix the one-off case of job_id not being populated unless annotation was completed, effectively meaning "ann_is_true" was forced on, but without actually using that flag, which is of course bad.
3. Fix a variety of > instead of >= and refactor to avoid in future. 
Note: It's possible this could have downstream effects, but unless there was some reason (I can't think of one) that we actually did want ">" for these, then this fix should still be correct, and we can do further fixes for downstream issues.
4. Small fixes on export conditions
5. Properly call `check_export_permissions_and_status()` in the immediate return case
6. Add a wip test case for this path since it's different from the existing working dir test case

A past hesitancy with moving to task abstraction was if we should actually export the task objects. But for now, by simply converting back to the same file list structure, it works. And we can explore exporting more task specific data later if needed.

Pablo, the test case mostly works but for some reason `instance_list` is not being populated. It's successfully getting the file though. https://github.com/diffgram/diffgram/runs/3680036283?check_suite_focus=true#step:17:109 
I'm not sure if I'm using the utility function correctly, maybe can I get your help there when you can please?

### DIFF
--- a/default/methods/source_control/file/file_browser.py
+++ b/default/methods/source_control/file/file_browser.py
@@ -625,7 +625,7 @@ class File_Browser():
                 limit = self.metadata["limit"],
                 offset = self.metadata["start_index"],
             )
-            if not working_dir_file_list or len(log['error'].keys()) > 1:
+            if not working_dir_file_list or regular_log.log_has_error(log):
                 return False
             file_count += count
         else:
--- a/default/methods/task/task_template/job_new_or_update.py
+++ b/default/methods/task/task_template/job_new_or_update.py
@@ -254,7 +254,7 @@ def job_update_api(project_string_id):
             project=project)
 
         job, log = job_update_core(session, job, project, input, log)
-        if len(log['error'].keys()) > 1:
+        if regular_log.log_has_error(log):
             return jsonify(log=log), 400
         log['success'] = True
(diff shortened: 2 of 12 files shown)


PJEstrada:
Hey @anthony-sarkis ! Change look great! The `log_has_error` function is a great idea! Is a lot more verbose and easier to understand. All other changes to the export look good to me too!

I'll take a look at the failing test case now and commit the fix

anthony-sarkis:
Thanks Pablo!!

PJEstrada:
Tests should now be fixed Anthony. Issue was just that the export mock was missing the `file_comparison_mode` field

PJEstrada (approved):

anthony-sarkis:
Amazing! Good find Pablo!!

anthony-sarkis:
Thank you very much!

PJEstrada:
Pleasure Anthony! I think with this we can merge it once tests pass :)

PR #308, opened 2021-09-29
Memory Guards, Queue Pressure Guard, Use cpu count to determine threads

PJEstrada:
To improve a bit video write performance

### DIFF
--- a/frontend/cypress/support/commands.js
+++ b/frontend/cypress/support/commands.js
@@ -200,6 +200,7 @@ Cypress.Commands.add('registerDataPlatformTestUser', function () {
   cy.wait(500);
   // cy.get('[data-cy="error-email"]').should('not.be.visible');
   cy.wait(3500);
+  cy.wait(2000);
   cy.url().should('eq', 'http://localhost:8085/user/builder/signup')
   cy.wait(2000)
   cy.get('[data-cy=first_name]').click();
--- a/walrus/methods/input/process_media.py
+++ b/walrus/methods/input/process_media.py
@@ -1107,15 +1107,20 @@ class Process_Media():
          of frame processing time (which is 25->40% of overall time.)
 
         """
+        print('Processing Frame: {}'.format(self.frame_number))
+        try:
+            result = self.read_raw_file()
+            if result is False: return False
 
-        result = self.read_raw_file()
-        if result is False: return False
+            self.process_image_for_frame()
 
-        self.process_image_for_frame()
-
-        if self.frame_number == 0:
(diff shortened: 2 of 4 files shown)


anthony-sarkis:
## Memory guards
Now system will check available memory and wait until memory is available to proceed
https://github.com/diffgram/diffgram/pull/308/commits/12a4137453e2801c60f1f9d232dad37fa7e4c099

Added it in
1) Getting initial item 75%
2) Main entry 75%
3) Video load 75%
4) Video write 85%
5) Frame write (every 10th frame) 90% limit

### Contexts:
1. memory needed in current setup is video size to download it, load it, plus the frames.
2. gunicorn worker context, a single node may have **5** workers, meaning that even limiting to a single video per thread is actually 5. This means the max video size becomes machine memory / 5. 
3. In many cases this is not relevant but in some larger video cases it causes a memory crash or over allocation request to kubernetes.
4. helps us move away from guessing mapping threads to expected load. And more towards actually checking what resources are available.
5. out of scope of retry system, since it's not actually an application issue directly, but the infrastructure lacking memory.

Example (was testing with smaller percent to trigger it)
memory warning

## Queue Pressure Guards
https://github.com/diffgram/diffgram/pull/308/commits/988ef9754005f716daa26b9678d049d7c67c9f40
The frame pressure especially can be quite large, if each is 3 MB uncompressed, than 3000 frames is 9 GB.
This assumes that the threads processing the queue can't keep pace.

Now, we guard against that and limit pressure to 2x queue size. During testing this appears to keep memory relatively flat.

frame processing queue pressure limit
protected memory

(After screenshot, commented out that logger for now, not sure what level we want that at)

## Num CPU cores
Great addition from Pablo as per his comment above. It does appear to process faster now so perhaps this was fixed in movie py :)
Updated a divide by 2 for guinicon
Also noting that the faster we process this it appears to reduce memory bloat - perhaps the memory bloat may in part be when the CPU can't reasonably write the video.

anthony-sarkis:
@PJEstrada I will await your review on this before merging :)

PJEstrada (commented):
Hey Anthony!,

All the code looks great! It's super good that the memory now stays stable most of the time :) that's great news!

I reviewed the code and both the memory and the queue pressure functions look good to me. Also the rest of there is nothing that stands out to me so if your are OK we can merge :)

This is something that will make the system sooo much more resilient! I really happy that you got those ideas on the memory checks.

anthony-sarkis:
Ok great! Thanks Pablo 😊

PR #309, opened 2021-09-29
Snap to Instance and Keypoints improvements

anthony-sarkis:
Main work is an addition of this snap to instance feature described here:
https://diffgram.readme.io/docs/snap-to-annotation
There's a couple of rough edges but in general it appears to be working as expected.
snap to another

Other items
1. Fixed generate samples -> was just missing filename for some reason
2. Increase zoom max -> user request, no clear reason we had it at number before
3. For Keypoints, option to have occluded keypoints hidden completely from view. This essentially makes them "deleted". and still leaves option to "undelete" them.
4. For Keypoints, enable mark point as _not_ occluded
5. Debugged that control key sticking issue a bit. Essentially the left control key seems to "stick", eg the keyup event never gets trigged. This seems to be some kind of odd ongoing issue.
http://jsfiddle.net/vor0nwe/mkHsU/   - easy way to show that the "up" event never fires
https://stackoverflow.com/questions/62683548/why-does-shiftleft-not-trigger-a-keyup-event-while-shiftright-is-held-and-vi    
We may want to think about using shift key instead or see if there's some more general work around. This started to feel like a time sink so left it for now.

Also:
1. refactored some of the hotkey stuff to try and make it a little more clear
2. started a concept of a `selected_instance_list`. a start to try to unify the various "selected" concepts.

### DIFF
--- a/frontend/src/components/annotation/annotation_core.vue
+++ b/frontend/src/components/annotation/annotation_core.vue
@@ -6595,7 +6595,7 @@
         },
 
         may_toggle_show_hide_occlusion: function (event) {
-          if (event.key === "o") {
+          if (event.key === "o" || event.key === "O") {
             this.label_settings.show_occluded_keypoints = !this.label_settings.show_occluded_keypoints
             this.refresh = new Date();
           }
--- a/frontend/src/components/annotation/annotation_core.vue
+++ b/frontend/src/components/annotation/annotation_core.vue
@@ -6574,6 +6574,26 @@
           }
         },
 
+        may_toggle_file_change_left: function (event) {
+          if (event.keyCode === 37 || event.key === "a") { // left arrow or A
+            if (this.shift_key) {
+              this.change_file("previous");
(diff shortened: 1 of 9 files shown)


PJEstrada (approved):
Hey Anthony! 

Al these changes look great!

Great find on the ctrl key thing. I found this related: https://www.mutuallyhuman.com/blog/keydown-is-the-only-keyboard-event-we-need/

Maybe we have to start thinking about this event with just the mouse down and setup the logic around that event only?

The snap to instance is also super cool! It's also going to be useful for the issues system.

I feel all the code look great Anthony! We can merge if you feel ready too :)

anthony-sarkis:
Thanks Pablo! And yes that's a good point about doing it all from the mouse event, then we perhaps wouldn't need to track the control key down at all - and therefore avoid having to mark it up again!

Yes great point for issues system too!! 

Ok great! I'll merge this. :)

PR #334, opened 2021-10-01
Drag keypoints fix

PJEstrada:
Fixes misplacement of keypoints when dragging entire instance


PJEstrada:
This branch should be ready now.

Main problem was that the drag_instance method was getting called twice. We already had the Interaction generator class calling it, so it was unnecesary to call it again in the move_something() function..

Also, the dragging was not considering the rotation. Had to change the math a bit to support dragging rotated points.

anthony-sarkis (approved):

PR #335, opened 2021-10-01
Keypoint label names DRAFT 

anthony-sarkis:
## Left Right Arrows

Now optional arrows show the left/right ness of a keypoint figure.

1. Arrows show left right even if moved - based on original template definition
2. Arrows smoothly rotate with rotation, eg a right point still shows as right even if rotated, and arrow stays correctly orientated. 
3. Dynamically adjust size

left right arrows all

Label names (strings) will come later

### DIFF
--- a/frontend/src/components/annotation/annotation_core.vue
+++ b/frontend/src/components/annotation/annotation_core.vue
@@ -33,7 +33,7 @@
                    :instance_type_list="instance_type_list"
                    :view_issue_mode="view_issue_mode"
                    :is_keypoint_template="is_keypoint_template"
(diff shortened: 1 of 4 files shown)


anthony-sarkis:
Great work @PJEstrada thank you very much for figuring out the core issue with the transformation reset!!

PJEstrada:
Link to Docs on Mouse Tools: https://dash.readme.com/project/diffgram/v1.0/docs/canvasmousetools

anthony-sarkis:
Another benefit is now the zoom percent correctly displays better when it's less then 100%. Before this was available through the panel size but was effectively hidden visually.

image

PR #356, opened 2021-10-11
Improve drag UX for keypoints instances

PJEstrada:
Now we can drag the instance without selecting the points, we also have 2 different pointer types for when a node is selected vs when the entire instance is being dragged.

https://www.loom.com/share/cd46e1cf26c34bd9b7bf5fb5597cfa5a

### DIFF
--- a/frontend/src/components/annotation/annotation_core.vue
+++ b/frontend/src/components/annotation/annotation_core.vue
@@ -2740,7 +2740,8 @@
 
           let index = update.index
           if (index == undefined) { return }  // careful 0 is ok.
-          const initial_instance = {...this.instance_list[index]}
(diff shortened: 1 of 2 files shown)


PJEstrada:
Hey @anthony-sarkis ! 

I pushed some fixes to make e2e tests pass on this one.

I just noticed one small thing with the hovered algorithm you proposed:

https://www.loom.com/share/ea07ae42a0b649b391c7a16fdb8d7edf

You can see there when I move a bit the mouse inside the node, there are some positions where the hover is not detected. This happens only when there is more than one node in the keypoint instance. Maybe the square root error is not detecting the node for some cases?

PR #359, opened 2021-10-12
Graceful Shutdown

PJEstrada:
### Main Changes
- Added `GracefulKiller` class to catch SIGINT AND SIGTERM signals.
- Refactored the thread startup process to be on a separate class and be done on demand in `main.py`, this prevents threads from being recreated on every import of process_media.py
- Added `ProcessMediaQueueManager` class to manages all the queues of the worker and centralize the queue management actions.
- Small visual improvement to the input list so that errors can be seen when clickin on the error icon.

### Gracefull Kill Algorithm Overview:

We will check the length of the `FRAME_QUEUE`, `VIDEO_QUEUE` and `PROCESSING_INPUT_LIST` and wait for a max of 2 hours until they are empty. If they all get empty, we kill the walrus. We also set the flag`STOP_PROCESSING_DATA=True` to avoid processing new messages on the current worker.

If more than 2 hours pass and all the messages are not handled, we set the messages status to `failed`.

## Notes:

This PR should be deployed alongside with this: https://github.com/diffgram/diffgram-helm/pull/5 to allow kubernetes to give a grace period for shutfdown on new deploys

There's still an issue when killing the walrus and a video file is getting written using moviepy. I've created an issue on their repo to ask about ways to gor around this: https://github.com/Zulko/moviepy/issues/1672

### DIFF
--- a/walrus/methods/utils/graceful_killer.py
+++ b/walrus/methods/utils/graceful_killer.py
@@ -4,7 +4,7 @@ import signal
 from walrus.methods.regular.regular_api import logger
 from shared.helpers.sessionMaker import session_scope
 from shared.utils.singleton import Singleton
-from signal import signal, SIGPIPE, SIG_DFL, SIGINT, SIGTERM, SIG_IGN
+from signal import signal, SIGINT, SIGTERM, SIG_IGN
 from shared.regular import regular_log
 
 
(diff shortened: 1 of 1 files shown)


PJEstrada:
Hey @anthony-sarkis!

This PR should now be ready to go. Please let me know if you find any issues with it or if there's something you think we need to change before meging

PR #400, opened 2021-11-01
Actions And Accounts Features From EE

PJEstrada:


PJEstrada:
Hey @anthony-sarkis !

I think all of the features should me now migrated. Please let me know if you find anything missing here. Otherwise we can just wait for tests and merge this now :)

anthony-sarkis:
Hey Pablo, **great work bringing this all together!!!**  😊 

A quick summary of what I changed
1) Worked on menu merge conflicts
https://github.com/diffgram/diffgram/commit/8f4ca621294c41d2202fa5c1f5e17c3e81e6600b
I removed old comments etc
2) marked some files as legacy
https://github.com/diffgram/diffgram/commit/8b19dc2bae518848734f6c902f537b37e8ae3961
3) deleted some files
https://github.com/diffgram/diffgram/commit/29839de7b9d86cdb3e651e7f062b71e0ef841b59
4) added the billing check. for now it's based on stripe billing (so not firing in open source)
This can be a good hook for further checks here eg the user thing etc.
https://github.com/diffgram/diffgram/commit/449aa5dbc84be61df1931cf5a88c370e344bc8cc
In that same line of thinking, I'd like to make the "diffgram.com" version stuff more clear.
I feel like there's a potential for someone to misinterpret the various billing things, and we need be clear 
that's only in the context of diffgram.com. (maybe even on docs or something too I'm not sure).

5) Made email work in new context
https://github.com/diffgram/diffgram/commit/fe7b370fad354260c353520e77743eae3c190067

Some other little things and links etc. like
https://github.com/diffgram/diffgram/commit/1a1bb05d90d0601e19ce2c5cbc05a827ea05b35b

I did some manual tests of links and things and in general things seem to be working as expected.

<br>

I think we should be prepared for some fast follow fixes since this is still a big merge, some of the test coverage is not ideal as we know etc.

The main question I'm not sure on, and we can discuss offline if needed, is the Almebic DB stuff. I do see you have the migration scripts there which is great!! 
I'm not quite sure what our process should be going forward. Do we just need to reset an existing database to the current head thing and then it will fall in line so we only have to do one migration thing?

anthony-sarkis (approved):

PJEstrada:
Hey Anthony! Thanks a lot for the review! 

- Already solved the failing test isssue thanks for solving the merge conflicts there!
- For the alembic migrations, yes I think for this one we can just move the head to be the latest version without actually doing a migration since the DB changes are already there.
- For the Stripe billing I agree with you, I think with the feature flag system we should be able to define it in a way where users should not even be aware of those things existing (unless they start diving into the source code). But I think the feature flags should be the longterm solution to this. I created a small entry in the docs for the review which I'll send later when ready

PR #401, opened 2021-11-01
Add mailgun variables request

vitalii-bulyzhyn:
In this branch:

- request Mailgun credentials on the installation process
- render login UI conditionally
- improve user experience


PJEstrada:
Hey Vitali!

Most of the code seems good to me, the only change I would suggest here is related to using the ENV variable in the frontend. This is somethin would not want for the following reasons:

- We don't want to expose that mailgun API key to our frontend systems for security reasons.
- On production we usually have the frontend service on a separate virtual machine as the backend services and potentially with different environment variables.
- We want to centralize the settings on single place in the backend and then make then front consume those settings via an API endpoint.

So  for now, for the sake of time, I think it's easier to keep the magic link button visible let's keep it that way and just make the python endpoint return an error response that we can display telling the user that the mail API key is not set.

For example:

```
Response:
Status: 400
{
"mailgun_api_is_set": false,
"message": 'The mailgun key is not set yet, please set MAILGUN_KEY"
}
```

A longer term solution would be to create a new endpoint on the backend like:

`/api/v1/settings/check/mail-api-key`

We call this during the mounting of the login screen, and based on the response of this we show/hide the magic link button.

If you think you can implement the above solution it's good too, however you feel more comfortable, I just feel that the first one might be a bit faster to implement. But I'll leave that choice to you.

The other thing is that whatever other things you encountered in the process of setting up the keys, please add them to the mail gun docs section here: https://diffgram.readme.io/docs/open-core-full-installation-mail

Let me know if I was clear enough with my comment here and as always feel free to ask me any other questions on any details you want

default/methods/configs/mailgun_is_set.py
```diff
@@ -0,0 +1,10 @@
+from methods import routes
+from shared.settings import settings
+
+@routes.route('/api/configs/is-mailer-set')
+def mailgun_is_set():
+    is_set = settings.MAILGUN_KEY and settings.EMAIL_DOMAIN_NAME and settings != 'localhost'
+    
+    if is_set:
+        return "true"
```
PJEstrada, line 9: I think here you could  use the `jsonify` method for flask, as we always assume that the server responses will be in a valid JSON format. so you could add something like `{"result": true}` just to follow the convetion on sending JSON as responses always.

default/routes_init.py
```diff
@@ -125,7 +125,11 @@ def do_routes_importing():
 
     from methods.ui_schema.ui_schema import ui_schema_new_api
 
+<<<<<<< HEAD
```
PJEstrada: I think there's a merge conflict here.

frontend/tests/unit/user/login.spec.js
```diff
@@ -40,4 +42,68 @@ describe("login.vue", () => {
     const showedMessage = wrapper.text();
     expect(showedMessage).toMatch(message);
   });
+
```
PJEstrada, line 45: These tests look great! :)

PJEstrada (changes requested):
Hey Vitali! I think code is 99% there just a couple things I pointed out in the commend and I think we can merge this one :)

default/methods/configs/mailgun_is_set.py
PJEstrada: In this case I think you'll also have to modify a bit your frontend methods to receive the value in the new format.

vitalii-bulyzhyn: Of course, will do that

install.py
```diff
@@ -413,10 +416,15 @@ def populate_env(self):
             env_file += 'POSTGRES_IMAGE={}\n'.format('tianon/true')
             env_file += 'DATABASE_URL={}\n'.format(self.database_url)
 
+        if self.mailgun:
+            env_file += 'MAILGUN_KEY={}\n'.format(self.mailgun_key)
+            env_file += 'EMAIL_DOMAIN_NAME={}\n'.format(self.email_domain)
```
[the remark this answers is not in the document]
vitalii-bulyzhyn: Oh yeah, forgot about containers

vitalii-bulyzhyn:
@PJEstrada just fixed everything you mentioned

PJEstrada:
> @PJEstrada just fixed everything you mentioned

Awesome thanks Vitalii! I will wait for the e2e tests to pass and if all is good we can merge this one :)

vitalii-bulyzhyn:
@PJEstrada I saw your commit, you are trying to press type password button, but there is no that button if mailgun is not set

PJEstrada (approved):

PR #406, opened 2021-11-02
Task Events

PJEstrada:
Task Events concept to move in the direction for storing a task history and removing the need of having a task type, so we can just manage the flows with a status column


PJEstrada:
Hey @anthony-sarkis !

I think this branch should be ready now. This is a list of the main changes:

- Added the relation columnd to the Job_to_User table.
- Added the TaskEvents table 
- Refactored any status changes to use `Task_Update` class so we can manage the event generation from there.
- Refactored the `complete_task()` function to check for review mode and request review if flag is on.
- Added all the task event generations on the `Task_Update` class
- Added the TaskUser table to allow reviewers and multiple assignees as well as the add review endpoint.
- Unit tests for all of the added functions.

I'm leaving out of this PR the review frequency support as it was getting pretty big already. So at this state we can consider that once the flag for review is on we have a 100% review chance for the tasks. I will also work on the reviewer assignation process (autobalancing reviewer assignation) on a separate PR too.

anthony-sarkis (approved):

Issue #403, opened 2021-11-01
Improve Task Status Reporting


PJEstrada:
When users go to the task template dashboard they still don't have good ways to monitor QA proceses and work in progress processes. We want to offer ways to easily answer questions suchs as:

- How much work is pending?
- How much tasks need to be reviewed?
- How Fast / Slow are we working?
- Any other related statistics reports that could help get a quick overview of the daily annotation operations.

@vitalii-bulyzhyn  will be leading this task with the context of the supporting document provided in the call.

PR #407, opened 2021-11-02
Add status to task report

PJEstrada:
Now we can group by status. This will be useful for reports needed in other contexts like : https://github.com/diffgram/diffgram/issues/403


anthony-sarkis (approved):
Great work Pablo!!!

anthony-sarkis:
The QA slideshow test was fixed on a nearby pull request so since that's only test (and appears to just be a testing bug not actual system) going to merge this!

PR #409, opened 2021-11-03
Add stats element

vitalii-bulyzhyn:
This is a draft for the statistics on the tasks page


default/methods/task/stats/fast_stats.py
```diff
@@ -0,0 +1,55 @@
...
+        instances_of_tasks = []
+        task_list = []
+        for task in tasks:
+            task_dict = task.__dict__
+            task_dict.pop('_sa_instance_state', None)
+            task_dict.pop('job', None)
+            task_dict.pop('incoming_directory', None)
+            task_list.append(task_dict)
+
+        return_task_list = [task for task in task_list if task["assignee_user_id"] == int(user_id)]
```
PJEstrada, line 33: Good job figuring out the API endpoints Vitalii! Just for your information, future direction of this would be to integrate it with the report runner system (Instead of using task.list()) but this is great for the v1 of it. We can optimize the queries in the future.

frontend/src/services/jobStatsServices.js
```diff
@@ -0,0 +1,11 @@
+import axios from 'axios'
+
```
PJEstrada, line 2: This is a much better patter to handle the services. I'll make sure to start using this and let's try to migrate whenever we can any endpoint direclty using URL's to this wrapper functions.

frontend/src/components/stats/stats_panel.vue
```diff
@@ -0,0 +1,268 @@
...
+                  Getting data for
+                  {{
+                    items.find((item) => item.id === current_user_stat).title
+                  }}
+                  ...
+                </span>
+              </div>
+            </template>
+            <v-list>
+              <v-list-item v-for="(item, index) in items" :key="index">
```
PJEstrada, line 100: For this one I would check the `<v-select>` component from vuetify or our wrapper `diffgram_select.vue` I think it will reduce the amount of code for this section of the component.

PJEstrada (commented):
Hey Vitalii! Overall code looks really good. Thanks for figuring out the endpoints and making the backend support  for this one. 

Some comments on mostly UI/UX stuff:

1.
image

I think it would be nice to add that count and progress bar to your stats component, so it's displayed in a more colorful and bigger way. It kinda gets lost in the UI now that you have the stats component.

2.
For this one:
image

Maybe a better label would be "Pending" instead of "Undone"

3.
For the name selector: 
image


I would try to add the select icon so users can know it's a interactable element. Something like this one:
image

Notice how it has a border and a downward arrow, this helps a lot to know that this is a select box. I added a comment below on the related code and how you could use the <v-select> component instead of this 

4.
Also for this one:

image

I'm not sure if chartjs support this, but it would be nice to see the percentages of each slice of the pie to get a numerical idea of the status too.


Other from that I think this is looking really good Vitalii!

default/methods/task/stats/fast_stats.py
vitalii-bulyzhyn: Yeah, I tried with report runner but at this point it required report template

frontend/src/services/jobStatsServices.js
vitalii-bulyzhyn: Thanks

frontend/src/components/stats/stats_panel.vue
vitalii-bulyzhyn: Of course, thanks for the suggestion

vitalii-bulyzhyn:
> image
> 
> I think it would be nice to add that count and progress bar to your stats component, so it's displayed in a more colorful and bigger way. It kinda gets lost in the UI now that you have the stats component.

I'm not sure what do you mean with the first point @PJEstrada

PJEstrada:
> > image
> > I think it would be nice to add that count and progress bar to your stats component, so it's displayed in a more colorful and bigger way. It kinda gets lost in the UI now that you have the stats component.
> 
> I'm not sure what do you mean with the first point @PJEstrada

What I was trying to say is that I think the UI would look better if you place this:

image


In another card inside your statis component, so instead of having 3 boxes we could have 4 and one of those could be that count of tasks. Let me know if I'm explaining myself here.

PJEstrada:
2 Other small layout things:

image

If we could make the width of the component to be 100% so that we don't have those blanks spaces on the red arrows and both the task list and the graphs countainer have the same width.

Last one:
image

Just moving the button to align with the title of the job so that we have a better layout there.

PJEstrada (commented):
Hey Vitalii!

Code's looking good! Thanks for the unit tests there. I think the only thing missing is moving the test of the available tasks. The idea I had is something like this:

image

That way we keep all the statistics in the boxes you created and we remove the text

vitalii-bulyzhyn:
@PJEstrada but basically, the first pie chart is what you ask me to add

vitalii-bulyzhyn:
@PJEstrada should be ready to merge now

default/methods/task/stats/fast_stats.py
```diff
@@ -0,0 +1,25 @@
+from methods.regular.regular_api import *
+from shared.database.task.job.job import Job
+from shared.database.user import User
+from shared.database.task.task import Task
+from shared.database.annotation.instance import Instance
+
+@routes.route('/api/job/<job_id>/stat', methods=["GET"])
```
anthony-sarkis, line 7: @vitalii-bulyzhyn for permissions, here's one example https://github.com/diffgram/diffgram/blob/ba092c6f11aaab9b38fc45c358807a06fe29ff90/default/methods/task/task/task_list.py#L6 
can search that wrapper for more examples.  all routes usually need permissions - the prior `is-mailer-set` was a rare exception since it effected login

PJEstrada (changes requested):
Hey Vitalii,

I just reviewed the code:

1. Please check @anthony-sarkis  previous comment to add the permissions to route, as at current point the API endpoint would be open to public.
2. I've added: https://github.com/diffgram/diffgram/pull/409/commits/d85d961d6281452616375ffbb481fb37f84aaf26 to optimizie the stats query. You can see main difference is that I removed the forloop we talked about. This is just for you to undertstand the performances changes of the code:

For example:
We have 10,000 tasks.

In previous status this code: 

```
            all_the_tasks = query.all()
            for task in all_the_tasks:
                instance = session.query(Instance).filter(Instance.file_id == task.file_id).count()
                instaces_created += instance
```

 would result in 10,000 queries to the database, which could potentially stress out the DB server and would make the response slower due to the DB repeatedly doing a similar query, and network latencies.

In the new version:

```
        task_id_list = [task.id for task in all_the_tasks_in_job]
        if user_id is None:
            # Get all the instances created on the tasks. No user filter
            instances_created = session.query(Instance).filter(
                Instance.task_id.in_(task_id_list),
                Instance.soft_delete == False
            ).count()
        else:
            # Get all the instances created by the given user
            user = User.get_by_id(session, user_id = user_id)
            instances_created = session.query(Instance).filter(
                Instance.task_id.in_(task_id_list),
                Instance.soft_delete == False,
                Instance.member_created_id == user.member_id
            ).count()
```

 we build a task_id list on a forloop in memory which does not make any queries. Then we make 1 single query to get all the instances with those task_ids. This results in a better performance as we just do 1 DB query. Notice also how I also added the `soft_delete=False` to avoid querying the archived instances and the `member_created_id` to query only the instances created by the given user. Let me know if you have any questions on this code, I know this is one of the first queries on DB you create so I wanted to detail as much as possible the reasons for the changes I made here.


I'm going to mark it with changes requested, just to make sure we add the endpoint permissions as this is an important thing to add before merging. Once that's ready we should be good to go here :)

vitalii-bulyzhyn:
@PJEstrada according to failed test: there is a problem, because it tries to press button but from what I see there were not enough tasks created in the test to have pagination

PJEstrada:
Hey @vitalii-bulyzhyn ,

I;ve checked those tests and it seems to have a method to increase the number of tasks:

`cy.createSampleTasksUsingBackend(15)`

Can you try increasing that to 30 maybe? I think maybe that could fix it

vitalii-bulyzhyn:
@PJEstrada of course, lemme try

anthony-sarkis:
Thank you for the optimization Pablo!

PR #411, opened 2021-11-03
Dynamic page titles

anthony-sarkis:
### Before 
every page was `diffgram`
image

For power users, demos, debugging etc. etc. it would be great to use vue router to change this

### After
image


PJEstrada (commented):
Hey Anthony! 

This is a nice feature! It will give users at lot more context on where they are. I'm just noticing that after changing the router a lot of e2e tests broke down.

Not sure if this has to do with the router or not but that's the only thing that jumped out to me on thi one

anthony-sarkis:
Thanks Pablo! 
I fixed the immediate issue here https://github.com/diffgram/diffgram/pull/411/commits/10d96ed33da21c444086fe75ff73f2d56b234f5a
and a related issue here https://github.com/diffgram/diffgram/pull/411/commits/9073b6443d8e7ed0a3ec4230901a24b9ea272646

In retrospect from working with this more I'm not certain if this is the only place we want to set the title. 

For example once we have loaded a job name, it would probably make more sense to have the name instead of the ID. in that case we might as well call `document.title` from the component. I think it's still good to have the default starting point here though, it makes it cleaner to edit static tiles, and the "default" dynamic titles that aren't set from a special condition.

This also ended up being a great chance to make that navigation guard way more readable and prep for router v4 upgrade.

it passes the key part of test on local and just waiting for CI to pass

PR #417, opened 2021-11-05
Paginate tasks

anthony-sarkis:
image

Adds basic pagination
fixes some other related things noticed during work


PJEstrada (approved):
Thanks for this one Anthony! The pagination will give a great perfomance boost to that area of the site :) I've checked all the code and it seems all good to me! I think just the new e2e test are pending and it should be good to got!

anthony-sarkis:
Tests passing! 🎉

PR #420, opened 2021-11-05
Fix autoborder

anthony-sarkis:
1) Actual bug was super tiny, single condition here https://github.com/diffgram/diffgram/pull/420/commits/00e61306d95fdac31eaf24dc8bdfb3ea5d2d28ab

2) Most of this effort is on adding the e2e test for autoborder and associated refactors

[test](https://github.com/diffgram/diffgram/pull/420/files#diff-56d541fcb1946fd116c98e756503ab8386323987d743685a4e7e4a39fb331fcd)

[refactor](https://github.com/diffgram/diffgram/pull/420/files#diff-c4906e16d3e24724e4cddf79cf7a4e37100d6d04aa6dcc3251374fcb89b2211cR293)

The new test may still need some little details otherwise this is a small one to merge

### DIFF
--- a/frontend/cypress/integration/diffgram/annotate_files/annotate_files_instance_drawing_spec.js
+++ b/frontend/cypress/integration/diffgram/annotate_files/annotate_files_instance_drawing_spec.js
@@ -81,50 +81,8 @@ describe('Annotate Files Tests', () => {
           {x: 200, y: 25},
         ]
 
-
-        cy.mousedowncanvas(points[0].x, points[0].y);
-        cy.mouseupcanvas()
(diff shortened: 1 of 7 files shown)


PJEstrada (approved):
Hey Anthony! Thanks for catching this one! code looks good to me. Also thanks for adding the commands for drawing the polygons that is going to be super useful for future tests and makes them much more readable :) I don't see anything that would prevent from merging this so I think we should be good to go.

anthony-sarkis:
Awesome! Thanks Pablo! :)

PR #426, opened 2021-11-08
Auto Assign Reviewer on Task Completion

PJEstrada:
Adds the reviewer with the least amounts of tasks automatically as a reviewer when a user completes a task and the job supports reviews.

### DIFF
--- a/default/methods/task/task/task_review.py
+++ b/default/methods/task/task/task_review.py
@@ -28,13 +28,7 @@ def task_review_api(task_id):
         return jsonify(log = log), 400
 
     with sessionMaker.session_scope() as session:
-        user = User.get(session)
-        if user:
-            member = user.member
-        else:
-            client_id = request.authorization.get('username', None)
-            auth = Auth_api.get(session, client_id)
-            member = auth.member
+        member = get_member(session = session)
 
         task_serialized = task_review_core(session = session,
                                            task_id = task_id,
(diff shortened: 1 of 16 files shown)


anthony-sarkis:
Hey Pablo, really great work on this! These new concepts and relations will open up a whole new path for human workflow.
I really like the volume of test coverage too!! I think the foundations overall look good to me. I saw a super tiny thing we have a function for we may be able to reuse, other then that I think it's good to merge. If you want do discuss any items in depth too we can do that :)   Best Anthony

default/methods/task/task/task_review.py
```diff
@@ -0,0 +1,90 @@
...
+        }}
+    ]
+
+    log, input, untrusted_input = regular_input.master(request = request,
+                                                       spec_list = spec_list)
+    if len(log["error"].keys()) >= 1:
+        return jsonify(log = log), 400
+
+    with sessionMaker.session_scope() as session:
+        user = User.get(session)
```
anthony-sarkis, line 31: We have the `member = get_member(session=session)` option.

I believe with no extra import, already available through regular_api. It provides a central point for changes and reduces code volume. one example https://github.com/diffgram/diffgram/blob/096f7628bd150cc958b084d0347dc501b79c07b8/default/methods/event/event_create.py#L56

I think that pattern is in 3 places on this pull request at least as far as git file search is showing searching `client_id = request.authorization.get('username', None)`

PJEstrada: Hey Anthony! Thanks for the commets! Sure I think this is a better way. I'll refactor those now.

PR #437, opened 2021-11-11
Handle a case with resetting the name

anthony-sarkis:
Context that we want a local copy for being able to edit the name
but we aren't getting the job data from this component
Not saying this is ideal but should work for now


anthony-sarkis:
This is a small bug fix - planning to merge once tests pass

PR #439, opened 2021-11-15
Add reviewers step to the task wizard

vitalii-bulyzhyn:


PJEstrada (commented):
Hey Vitalii!

Thanks for the code, the code seems to be going in a good direction!  

here are some comments:

1. Text is duplicated in the first title:
image

2. I would try to change the text input for the review chance into a `v-slider` as with the current UI, it's not clear what type of values the user has to enter, and would save you the time of validating the text input is in the correct format. Also we can default this value to be 1.0 (100%) by default so that by default all tasks are reviewed. Also make sure to only allow float values between 0 and 1, as the backend should only accept those values.
image


3. The UI is not saving anything yet, if I updated the review chance and set the value of reviews to "yes" the changes are not reflected on the DB. Make sure to update the `allow_reviews` and `review_chance` of the `/api/v1/project/<string:project_string_id>/job/update` route. You can test this by setting the values and then closing the browser tab and returning to the draft job. The values you set should be saved there.

4. Last thing is that it seems that some e2e tests broke from the changes on the steps. Please make sure to fix those and add the logic to test this new step that was added. Also make sure to add some unit tests to the new components.

As a more general note, we always try to test the UI as an end user to make sure the new feature is working in the context of the entire system like backend saving, testing the rest of steps in the wizard etc. While doing that, I try to ask the question: Will the end user feel this is working as expected? That mindset can help a bit to find some functionality bugs that might not be so obvious when reading at the code directly. 

Other from that I think code is going in really good direction! 

Let me know if you have any other questions, or if you need any help with any of the points :)

vitalii-bulyzhyn:
@PJEstrada about the 3rd one - yes, it's not wired up yet. Was just a UI. I need a bit of the comments on the saving part on the call today
and 4th - same, since it's just draft I didn't fix tests just yet

PJEstrada:
> @PJEstrada about the 3rd one - yes, it's not wired up yet. Was just a UI. I need a bit of the comments on the saving part on the call today and 4th - same, since it's just draft I didn't fix tests just yet

Oh ok cool no problem, we can discuss the detail there

default/methods/task/task_template/job_new_or_update.py
```diff
@@ -542,7 +564,9 @@ def new_web(project_string_id):
             reviewer_list_ids=input.get('reviewer_list_ids'),
             attached_directories_dict=input['attached_directories_dict'],
             pro_network=input['pro_network'],         
-            default_userscript_id=input['default_userscript_id']
+            default_userscript_id=input['default_userscript_id'],
```
PJEstrada, line 567: Thanks for adding this Vitalii!

PJEstrada (commented):
Hey Vitalii! The code is looking pretty neat :) thanks for all the work here.

I just a have  1 super small thing to add here:


When going to a job that was previously saved in draft mode, and the checkbox "review all" was checked. It does not seem to save that state of the checkbox. Although the 100% is saved, I supsect it might just be a rendering issue. Here are the steps I took to replicate it:

1. Create a new task template
2. Go to reviewer steps, click "yes" option.
3. Click the review all check box.
4. Close the tab or got to the job list section.
5. Click on the job that was previously working on (should be on draft status)
6. Go to review step, you will notice the checkbox is not checked, but the value of the slider is 1.

Other from this I think we should be good :)

PJEstrada (approved):

PR #443, opened 2021-11-16
Task completion dialog

vitalii-bulyzhyn:


PJEstrada (commented):
Hey Vitalii! 

Code looks super good. Really like the direction is taking. Some UX comments when I tried using it.

1. This is related to the first dialog (`review_dialog`). I'm not sure if we should display this because in some cases they might hit submit to review and the task would not actually get reviewed because of the random chance. Maybe a better way would be if the task was actually sent, that we show a dialog or a snackbar after the request was completed just notifying the user if the task was actually sent to review or not.
2. The second dialog looks great! I think we can leave the textbox always available because sometimes users might just want to comment for props and encouraging good work in cases where no changes are required.

It can be similar to the github one I'm writing on:

image

3. You can check the isssues system we have on diffgram to add a rich text editor on the comment box. 
4. I know this is still WIP, but just to add to this, maybe we want to show the review status in the task list component as well as who is the reviewer.

I'm adding some extra comments on some files below.

frontend/src/components/annotation/toolbar.vue
```diff
@@ -1,282 +1,278 @@
 <template>
-<v-toolbar
-      v-if="show_toolbar"
-      dense
-      width="100%"
-      elevation="0"
-      fixed
-      :height="height"
-      style="overflow: hidden; padding:0; border-bottom: 1px solid #e0e0e0"
+  <v-toolbar
```
PJEstrada, line 2: For this file. It seems that mostly what changed was indents right? I think we can have a consensus on how do we set our editor to auto indent code so that we don't have issues where files seems to have changes and there are only indents. We can define it with something like: https://editorconfig.org/

vitalii-bulyzhyn:
@PJEstrada about #4 I just opened issue #445, I think I will do this on the separate PR

frontend/src/components/annotation/toolbar.vue
vitalii-bulyzhyn: Yeah, that totally has sense. We don't have any config file so far, right?

PJEstrada (commented):
Hey Vitalii!

Code looks great the new function `on_task_annotation_complete_and_save()` looks a lot better and makes it easier to understand the completion process vs the saving process.

I think we now just need the unit/e2e tests and this can be ready to test!! 

Please always make sure to do some manual testing of the entire feature before the final review to double check that any other parts are not breaking. 

Thanks a lot for this one Vitalii and Good work!

vitalii-bulyzhyn:
@PJEstrada thanks for the review

PJEstrada (changes requested):
Hey Vitalii, I just tested the branch. Here are some issues I had:

1. 
image
When trying to save an image the loading state stays stuck even though the server already responded 200.

2. I tried making a review and rejecting it:
image

But frontend seems to still send approve to the backend:

image

So the task ends up completed instead of in progress again.



Those are the only 2 bugs I found. It's always a good idea to do some manual testing as an end user to spot these types of issues. I always tend to create a new task template from scratch and go through the entire process when I finish writing the code to make sure everything is working as expected, testing al possible flows of the system. I think that will help you to spot those issues before the review for next time.

Let me know if you have any questions on these

vitalii-bulyzhyn:
@PJEstrada thanks for the review, lemme fix them

PJEstrada (approved):
Thanks for adding tests Vitalii!

We should be good to go now, feel free to merge this to master :)

PR #449, opened 2021-11-22
Display new statuses on job page

vitalii-bulyzhyn:
In this PR:
- Display new icons for new statuses on the task list
- Display new statuses on the stats

### DIFF
--- a/frontend/src/components/regular_concrete/task_status_icons.vue
+++ b/frontend/src/components/regular_concrete/task_status_icons.vue
@@ -38,6 +38,15 @@
   </tooltip_icon>
 
   <tooltip_icon
+    v-if="status == 'requires_changes'"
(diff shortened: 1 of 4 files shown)


vitalii-bulyzhyn:
@PJEstrada do you think I need to add extra units for this?

PJEstrada:
Hey Vitalii! If unit tests already exist I would just add any extra assertions to cover this code. If the component tests are non existent then I think its always good to add them. As long term goal we would like to eventually reach 90% tests code coverage so any steps towards that goal are always a good idea :)

vitalii-bulyzhyn:
Gotcha

PJEstrada (approved):
Hey Vitalii!

Thanks for the improvements here I see everything good now. Feel free to merge this one to master now

PR #451, opened 2021-11-24
Manual user assignment  to the task

vitalii-bulyzhyn:
This PR contains:
- manual assignment/removing of the user to the single task (reviewers and annotators)
- manual batch user assignment/removing (reviewers and assigners)
- conditional rendering of the reviewers column
- add docs on task users management: https://diffgram.readme.io/docs/task-user-management


default/methods/task/task/task_user_add.py
```diff
@@ -17,19 +18,26 @@ def api_task_user_add(project_string_id, task_id):
...
 
+        result = {}
+
         log, input, untrusted_input = regular_input.master(request = request,
                                                            spec_list = spec_list)
         if len(log["error"].keys()) >= 1:
             return jsonify(log = log), 400
+        
+        for user in input['user_id']:
+            users_already_assigned = session.query(TaskUser).filter(TaskUser.task_id == task_id).filter(TaskUser.relation ==  input['relation']).filter(TaskUser.user_id == user).count()
```
PJEstrada, line 29: Hey Vitalii!

Thanks for adding the support for lists here. As a small convention on the backend, we always try to move any logic to the core logic layer instead of the API function.

So in this example, what I would do is just move the user forloop inside the `api_task_user_add_core` function. That way we can leave in the API just the permissions and the payload type checking. 

In general we always do it in the following way:

1. They api function ( in this case `api_task_user_add`) has the responsability of checking permissions (with the decorators or any other specifics if necessary),  creating the DB session, checking the payload types are correct, and then calling the core function that will process the actual business logic.
2. The core function (in this case `api_task_user_add_core`) has the responsabiliy of executing the business logic and returning the serialized result so that the API function can return it to the client.

default/methods/task/task/task_user_add.py
```diff
@@ -17,19 +18,26 @@ def api_task_user_add(project_string_id, task_id):
...
         ]
 
+        result = {}
+
         log, input, untrusted_input = regular_input.master(request = request,
                                                            spec_list = spec_list)
         if len(log["error"].keys()) >= 1:
             return jsonify(log = log), 400
+        
+        for user in input['user_id']:
```
PJEstrada, line 28: for the `input['user_id']` I would just rename it to `input['user_id_list']` just to make it more explicit for other develiper that this key is a list.

default/methods/task/task/task_user_modify.py
```diff
@@ -0,0 +1,108 @@
+from methods.regular.regular_api import *
```
PJEstrada, line 1: This one looks great Vitalii, I think you now have the grasp of how the backend is structured.

The only comment is the same as with the previous, to keep the api and core logic separated. I created a small commit as an example of how would I refactor it here: https://github.com/diffgram/diffgram/pull/451/commits/ef3e1285c03bbf26e419f1a0b67b077f1637fa2d

You can use this one to make the refactor to the `task_user_add` API

default/methods/task/task/task_user_remove.py
```diff
@@ -3,23 +3,43 @@
...
         if len(log["error"].keys()) >= 1:
             return jsonify(log = log), 400
 
-        result, log = task_user_remove_core(
-            session = session,
-            task_user_id = task_user_id,
-            project_string_id = project_string_id,
-            log = log
-        )
+        for user in input['user_id']:
```
PJEstrada, line 29: Same as above comments. 

- Changing the `user_id` to `user_id_list`
- Moving the forloop logic into the core function to keep the API with just the above mentioned responsabilities.

default/tests/methods/task/task/test_task_user_add.py
```diff
@@ -60,7 +60,7 @@ def setUp(self):
     def test_api_task_user_add(self):
         request_data = {
             'relation': 'assignee',
-            'user_id': self.member.user_id
+            'user_id': [self.member.user_id]
```
PJEstrada, line 63: Thanks for updating the tests!

default/methods/task/task/task_user_add.py
vitalii-bulyzhyn: Got it

default/methods/task/task/task_user_add.py
vitalii-bulyzhyn: Yeah, makes sense

PJEstrada (commented):
Hey Vitalii! Thanks a lor for all the work on this, really good job overall!

I added a couple comments mostly on some backend conventions we have, but I see most of the code with no big issues really. 

I also already tested the feature and it's working fine :+1:  thank you very much!

PJEstrada:
One other small UX comment. For the removing of assignee/reviewers in bulk, It would be good that we could have a different text for the dialog since we are removing them. It was a bit confusing that when I clicked removed annotators the dialog then said: 

Who should be assigned to this task?

image

Or maybe another one could just to call the action, Edit Annotators/ Edit Reviewers, whatever option's easier

vitalii-bulyzhyn:
@PJEstrada can you check it one more time please, I refactored everything according to your advice

PJEstrada (approved):

PJEstrada:
Hey Vitalii! Thanks for your work in this. The PR should be ready now, let's just wait for tests and we can merge this one :)

PR #453, opened 2021-12-01
Enterprise migration and new admin display

anthony-sarkis:
Adds a basic install info display.
Pulls in memory variables. sensitive info is wrapped with `bool()`
In the future we could also consider editing some of these through this UI screen.
Context of debugging trailing slash issue (also solved in this request)

image

image


PJEstrada:
Awesome Anthony this is a great idea! We can eventually transform this into to a full admin dashboard and save values on DB and leave only the completely necessary (like DB url) as ENV variables.

PJEstrada (approved):

PR #458, opened 2021-12-08
fix: paste to next frames now fetches missing instance buffers

PJEstrada:
When pasting to next frames, if the selected next frames where not yet on the instance buffer dict, the pasted instance would overwrite the entire frame.

Now we fetch any missing frames beforehand to guarantee that the pasted frame  is added to the existing instance list.

### DIFF
--- a/frontend/src/components/annotation/annotation_core.vue
+++ b/frontend/src/components/annotation/annotation_core.vue
@@ -6942,31 +6942,39 @@ export default Vue.extend({
       const step_size = 5; // We will fetch 5 frames per call
       const limit = pLimit(15); // 10 Max concurrent request.
       const total_frames = frames_end - frame_start;
-
-      // Build frames start/end
-      const frames_tuples = [];
-      for (let i = frame_start; i < frames_end; i += step_size + 1) {
-        frames_tuples.push([i, i + step_size]);
-      }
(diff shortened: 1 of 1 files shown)


anthony-sarkis (approved):

anthony-sarkis:
Good work Pablo!! 

One question, error handling for the buffer function in general, https://github.com/diffgram/diffgram/blob/3e5faa4b0ca102fb8c263e0fc3d5f3e8d78cc8e1/frontend/src/components/annotation/annotation_core.vue#L6941

Maybe I'm just missing it but this existing function does not seem to throw a user facing error in case of failure. 
In the saving case, we may want to break off trying to save if it fails to load the frames

PJEstrada:
> Good work Pablo!!
> 
> One question, error handling for the buffer function in general,
> 
> https://github.com/diffgram/diffgram/blob/3e5faa4b0ca102fb8c263e0fc3d5f3e8d78cc8e1/frontend/src/components/annotation/annotation_core.vue#L6941
> 
> Maybe I'm just missing it but this existing function does not seem to throw a user facing error in case of failure. In the saving case, we may want to break off trying to save if it fails to load the frames

Thanks Anthony! I've already added better error handling to it to stop the pasting process in case of any problem. Thanks for cathing that, I'll merge now and deploy

PR #466, opened 2021-12-14
Feature checker & upgrade button

PJEstrada:
- Add feature checks on free tiers and a more visible upgrade button.
- Attach install fingerprint to orders so that open source installation can also be upgraded.
- Add class for feature flag checker.
- Add stronger constraints for hosted free plans


PJEstrada:
Hey @anthony-sarkis!

Now this popup will showup when a free tier limit is reached in diffgram.com

image


Open source installation will still have no limits.

Still pending some checks on the export and project settings, but wanted to show the mock if you want to add any changes or fix a bit text messages

PJEstrada:
Hey @anthony-sarkis ! This PR should be ready for review now :)

Main Features added:
- Added free limit dialog to display whenever there is a free limit reached
- Added blocks for images, videos, text files and 3d files uploads.
- Added blocks for frame count on video
- Added blocks for export instance count.
- Added FeatureChecker class to centralize feature flags and plan checks.

Some notes we can discuss:
- I removed the permissions on the route `new_plan_api` to allow for anonymous users to make a purchase as long as the `finger_print_id` and `email_opensource` is provided. We can discuss if you feel we can find another solution to this as I know it's not that common to have an open endpoint on the API. Let me know your comments on this :)

anthony-sarkis:
Great work Pablo!!! I think this will really help us grow Diffgram

Let's discuss on call tomorrow

walrus/methods/input/process_media.py
```diff
@@ -1026,13 +1031,66 @@ def process_sensor_fusion_json(self):
             self.input.status = 'failed'
             self.input.update_log = self.log
 
+    def check_free_tier_limits(self):
```
anthony-sarkis, line 1034: If open source or premium plan this should return before doing a SQL query right? Like in `check_export_billing`

On free plans, on the old method we cached the "last checked" thing, so for performance this query would only run once every x minutes or so. I think that's a good idea to keep if reasonable

shared/feature_flags/feature_checker.py
```diff
@@ -0,0 +1,123 @@
...
+            roi_annotation_productivity_benefit = -1,
+            roi_project_quality_improvement = -1,
+            roi_total_benefit = -1,
+            roi_multiple = -1
+
+        )
+        return plan
+
+    def get_flag(self, flag_name):
+        if not settings.ALLOW_PLANS:
```
anthony-sarkis, line 77: What I'm picturing for open source part

1. We have a really clear "is open source" flag
2. Then we have a clear "open source max user" variable (eg = 20)
3. Then all the various feature flag checkers

    1. eg if open source true and not user count, then returns without doing anything (allow all). In future if we need to add stuff here to prevent abuse we can.

    2. if it's user related, then it checks "open source max user" the user count, but that's it

If we can build this into the feature_checker level always then it can help make it clear we aren't limiting people in open source context.

Sorry in advance if this is already in the code and I missed it

So maybe `IS_OPEN_SOURCE`. defaults to True, and then for diffgram.com saas we can set to false?

Potentially `ENABLE_PLANS` maybe still is a good idea to have as a feature flag just in case there is say a bug in a plan and we want to disable it for customer service reasons etc.

walrus/methods/input/process_media.py
PJEstrada: Yes in this case it should run before the queries and main processing. But we do it after the upload, that's something we still have to do because we need to count frames for videos specifically.

shared/feature_flags/feature_checker.py
PJEstrada: Those are good ideas Anthony! Thanks for sharing

I was thinking in a similar way but not having the `IS_OPEN_SOURCE` I was thinking that all open source instalation will have the `ALLOW_PLANS` and `ALLOW_BILLING` as False with the exception of difffgram.com. That way all user install will never do any plan checks, but maybe the `IS_OPEN_SOURCE` is a bit more explicit. We can discuss more details on the call!

PR #489, opened 2022-01-04
Improve sequence list

anthony-sarkis:
### Main items
1. **Backend**: Extract keyframe cache to be more composable. This splits the old and confusing thing hidden in the `update` calls into much more clear stuff ->

image

The main concept here is to make it easier to understand when keyframes are getting added or removed from the cache

image
https://github.com/diffgram/diffgram/commit/83b635759511c9ab3d1541b65f3175db88093d5b  (plus some little updates after)

**Timing bug**
There was a timing bug here too - depending on the order of it it was possible that the deletion would overwrite the addition.
eg ->  added instance adds frame number 29 then the removed version of it goes and deletes it. it's correct if the order is inverted. but order is hard to setup here. 

Solution was to just tracking which sequences we already added to, and then using that as an ignore list.

I think overall this will still need more iterations over time, but at least now it's pulled closer to the decision point and should be a lot more workable.

**little fixes and performance boost**
Also fixed a variety of little cases with the inserts, eg before repeated calls could have resulted in more deletions etc.
AND performance boost, was able to get rid of the `in list` check by adding a bisect plus some conditions. hat should mean that for large keyframe lists it's binary search time instead of checking each element.
https://github.com/diffgram/diffgram/blob/61e34c951209e97ac10537f101d279616a2173cb/shared/database/video/sequence.py#L491  

2. **Frontend**: 
  a) Better conditions on when to add keyframes on front end. bug fix that sometimes it was adding all.
  b) Added a proactive removal method
https://github.com/diffgram/diffgram/commit/73f958ef1991b68e72b9d8636c7c8fea2758a33
adding 55s to all

The benefit of leaning on backend here is that it helps ensure alignment, but the downside is that it's a little clunky. 

I think it would be better to do this before save.

It seemed like it would break the pattern too much do that change now but something to keep in mind for v2 of the sequence browser. Really trying to move away from this dependency on the backend for stuff we already know on frontend. Rather have backend act more like a saving mechanism in this case then a blocker

## Related bug observation - question for Pablo 
The only state I was able to get into that would break this on backend was when the frontend failed to send instance list. eg by moving quickly in UI - some edge case of mashing buttons we haven't covered yet.

not having all the instances available can still cause keyframe issues

I'm suggesting we go back to hard block on this for now, since it also breaks the keyframe thing other stuff.
https://github.com/diffgram/diffgram/blob/61e34c951209e97ac10537f101d279616a2173cb/shared/annotation.py#L440

Happy to revert quickly if we get too many hits on it, but perhaps when we do we can try to button up the front end, maybe block moving around in certain saving, or just make it more robust to allow navigation while saving (I think we took some steps there but maybe not 100% yet?)

Please let me know thoughts on this especially

### Relying on the added/deleted and action_type
This stuff is starting to depend more on some of those flags and states. We may need some more iterations on that too.

### Other items and side notes
1. Identified regen as totally different case which is regen all from a fresh instance_list() pull. updated new function for that case -> https://github.com/diffgram/diffgram/commit/5b4677bb16798eae5be2dbe1c26b7ba5717c29c7
2. noting this tiny logging change here just in case I'm missing a reason we needed it https://github.com/diffgram/diffgram/commit/f007bbe716e8c2ecf132e6b6caf95f419abb65dd 

Edit:
The tests were passing but the new False thing broke them, I tried updating here
https://github.com/diffgram/diffgram/pull/489/commits/99fbb01d2d52e4fe7e41b9c9f0e9f35b3278664c

### DIFF
--- a/default/main.py
+++ b/default/main.py
@@ -53,6 +53,8 @@ def log_request_info():
 	try:
 		if request.path in do_not_log_these_routes:		# sensitive routes
 			return
+		if request.path.endswith("annotation/update"):
+			return
 		out = request.get_json(force=True)
 		out['path'] = request.path	# Because may not log in line with normal path printing.
 		logger.info(out)
--- a/default/tests/shared/test_annotation.py
+++ b/default/tests/shared/test_annotation.py
@@ -1196,7 +1196,7 @@ class TestAnnotationUpdate(testing_setup.DiffgramBaseTestCase):
         )
         result = ann_update._Annotation_Update__check_all_instances_available_in_new_instance_list()
         print('ann_update.log', ann_update.log)
-        self.assertFalse(result)
+        self.assertTrue(result)
         self.assertTrue(len(ann_update.log['warning'].keys()) > 0)
         self.assertTrue('new_instance_list_missing_ids' in ann_update.log['warning'])
         self.assertTrue('information' in ann_update.log['warning'])
@@ -1219,7 +1219,7 @@ class TestAnnotationUpdate(testing_setup.DiffgramBaseTestCase):
         )
         result = ann_update._Annotation_Update__check_all_instances_available_in_new_instance_list()
 
-        self.assertFalse(result)
+        self.assertTrue(result)
         self.assertTrue(len(ann_update.log['warning'].keys()) > 0)
         self.assertTrue('new_instance_list_missing_ids' in ann_update.log['warning'])
         self.assertTrue('information' in ann_update.log['warning'])
--- a/default/tests/shared/test_annotation.py
+++ b/default/tests/shared/test_annotation.py
@@ -1196,7 +1196,7 @@ class TestAnnotationUpdate(testing_setup.DiffgramBaseTestCase):
         )
         result = ann_update._Annotation_Update__check_all_instances_available_in_new_instance_list()
         print('ann_update.log', ann_update.log)
-        self.assertTrue(result)
+        self.assertFalse(result)
         self.assertTrue(len(ann_update.log['warning'].keys()) > 0)
         self.assertTrue('new_instance_list_missing_ids' in ann_update.log['warning'])
         self.assertTrue('information' in ann_update.log['warning'])
@@ -1219,7 +1219,7 @@ class TestAnnotationUpdate(testing_setup.DiffgramBaseTestCase):
         )
         result = ann_update._Annotation_Update__check_all_instances_available_in_new_instance_list()
 
-        self.assertTrue(result)
+        self.assertFalse(result)
         self.assertTrue(len(ann_update.log['warning'].keys()) > 0)
         self.assertTrue('new_instance_list_missing_ids' in ann_update.log['warning'])
         self.assertTrue('information' in ann_update.log['warning'])
--- a/frontend/src/components/annotation/annotation_core.vue
+++ b/frontend/src/components/annotation/annotation_core.vue
@@ -8111,6 +8111,7 @@ mplate_has_keypoints_type: function (instance_template) {
           if (this.video_mode) {
 
             for (var instance of response.data.added_instances) {
+
               this.add_keyframe_to_sequence(instance)
 
               if (instance.action_type == "deleted") {
--- a/frontend/src/components/annotation/annotation_core.vue
+++ b/frontend/src/components/annotation/annotation_core.vue
@@ -8112,12 +8112,12 @@ mplate_has_keypoints_type: function (instance_template) {
 
             for (var instance of response.data.added_instances) {
 
-              this.add_keyframe_to_sequence(instance)
+              this.add_keyframe_to_sequence(instance, current_frame_cache)
 
               if (instance.action_type == "deleted") {
                 this.$refs.sequence_list.remove_frame_number_from_sequence(
                   instance.sequence_id,
-                  this.current_frame)
+                  current_frame_cache)
               }
             }
           }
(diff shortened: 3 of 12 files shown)


PJEstrada:
Hey Anthony!

Thanks for those updates, this really makes the sequence saving a lot more robust :)

* For the questions on the empty instance lists being sent. I think it's a good idea to have a check for it but maybe we have to check only check if the file already has instances. If we don't check this it may conflict with files that have global attributes (and no instances) in future versions. I'm not sure if an instance is created behind the scenes with global attributes but that was the only case I could come up with where it may conflict a bit.

* For reactivating `__check_all_instances_available_in_new_instance_list` hard check I would really like to have that on eventually again, but I feel the video annotators tend to do this quick frame changes, copy/paste at really high speed so it will most likely affect the annotators workflow. Specially since I did a query for the `missing_ids_in_new_instance_list_error` event and we still have hits today (more than 1000) for it. Maybe we can explore a bit more how are you getting the buttons smashing to eventually lead to the issue so we can analyze the root cause and missing blocks on the frontend? Also it seems that we might be able to replicate it 

This is the hit count of the event since august 2021:

"2022-01-04 00:00:00"	1441
"2022-01-03 00:00:00"	311
"2022-01-02 00:00:00"	1
"2021-12-31 00:00:00"	12
"2021-12-30 00:00:00"	167
"2021-12-29 00:00:00"	325
"2021-12-28 00:00:00"	129
"2021-12-27 00:00:00"	131
"2021-12-24 00:00:00"	3
"2021-12-23 00:00:00"	68
"2021-12-22 00:00:00"	50
"2021-12-21 00:00:00"	20
"2021-12-20 00:00:00"	23
"2021-12-19 00:00:00"	47
"2021-12-18 00:00:00"	74
"2021-12-17 00:00:00"	143
"2021-12-16 00:00:00"	380
"2021-12-15 00:00:00"	175
"2021-12-14 00:00:00"	933
"2021-12-13 00:00:00"	203
"2021-12-11 00:00:00"	17
"2021-12-10 00:00:00"	171
"2021-12-09 00:00:00"	438
"2021-12-08 00:00:00"	581
"2021-12-07 00:00:00"	636
"2021-12-06 00:00:00"	566
"2021-12-04 00:00:00"	8
"2021-12-03 00:00:00"	34
"2021-12-02 00:00:00"	56
"2021-12-01 00:00:00"	42
"2021-11-30 00:00:00"	20
"2021-11-29 00:00:00"	36
"2021-11-26 00:00:00"	227
"2021-11-25 00:00:00"	9
"2021-11-24 00:00:00"	29
"2021-11-23 00:00:00"	5254
"2021-11-22 00:00:00"	20
"2021-11-21 00:00:00"	1
"2021-11-19 00:00:00"	17
"2021-11-18 00:00:00"	7
"2021-11-17 00:00:00"	63
"2021-11-16 00:00:00"	336
"2021-11-15 00:00:00"	1096
"2021-11-12 00:00:00"	151
"2021-11-11 00:00:00"	108
"2021-11-10 00:00:00"	10
"2021-11-09 00:00:00"	104
"2021-11-08 00:00:00"	41
"2021-11-05 00:00:00"	52
"2021-11-04 00:00:00"	7
"2021-11-03 00:00:00"	4
"2021-11-02 00:00:00"	2
"2021-11-01 00:00:00"	5
"2021-10-29 00:00:00"	15
"2021-10-28 00:00:00"	36
"2021-10-27 00:00:00"	87
"2021-10-26 00:00:00"	55
"2021-10-25 00:00:00"	17
"2021-10-24 00:00:00"	389
"2021-10-23 00:00:00"	103
"2021-10-22 00:00:00"	735
"2021-10-21 00:00:00"	75
"2021-10-20 00:00:00"	12
"2021-10-19 00:00:00"	208
"2021-10-15 00:00:00"	67
"2021-10-13 00:00:00"	4
"2021-10-11 00:00:00"	4
"2021-10-08 00:00:00"	11
"2021-10-07 00:00:00"	5
"2021-10-06 00:00:00"	6
"2021-10-05 00:00:00"	3
"2021-10-01 00:00:00"	4
"2021-09-30 00:00:00"	17
"2021-09-29 00:00:00"	29
"2021-09-28 00:00:00"	35
"2021-09-27 00:00:00"	26
"2021-09-24 00:00:00"	88
"2021-09-23 00:00:00"	1395
"2021-09-22 00:00:00"	1424
"2021-09-21 00:00:00"	1134
"2021-09-20 00:00:00"	16
"2021-09-18 00:00:00"	66
"2021-09-17 00:00:00"	83
"2021-09-16 00:00:00"	636
"2021-09-15 00:00:00"	158
"2021-09-14 00:00:00"	156
"2021-09-13 00:00:00"	28
"2021-09-10 00:00:00"	7
"2021-09-09 00:00:00"	8
"2021-09-08 00:00:00"	7
"2021-09-07 00:00:00"	1
"2021-09-06 00:00:00"	32
"2021-09-02 00:00:00"	1
"2021-09-01 00:00:00"	106
"2021-08-31 00:00:00"	269
"2021-08-30 00:00:00"	594
"2021-08-29 00:00:00"	30
"2021-08-28 00:00:00"	238
"2021-08-27 00:00:00"	149
"2021-08-26 00:00:00"	1215
"2021-08-25 00:00:00"	206
"2021-08-24 00:00:00"	67
"2021-08-23 00:00:00"	138
"2021-08-21 00:00:00"	11
"2021-08-20 00:00:00"	75
"2021-08-19 00:00:00"	648
"2021-08-18 00:00:00"	435

anthony-sarkis:
Hey Pablo, 
Thanks for catching that I think that's a great point, I made the revert on that. 
And made a seperate ticket here for that
https://github.com/diffgram/diffgram/issues/490
Best Anthony

PR #491, opened 2022-01-04
Fix Concurrent Saving Frames Issues

PJEstrada:
Improvements to prevent frames from not getting saved and wrong frame to be updated with new IDs when saving data.

### DIFF
--- a/frontend/src/components/annotation/annotation_core.vue
+++ b/frontend/src/components/annotation/annotation_core.vue
@@ -3462,7 +3462,7 @@ mplate_has_keypoints_type: function (instance_template) {
       }
 
       let keyframes_to_sequences = this.build_keyframes_to_sequences_dict();
-
+      console.log('keyframes_to_sequences', keyframes_to_sequences)
       this.populate_ghost_list_with_most_recent_instances_from_keyframes(
         keyframes_to_sequences
       );
@@ -3497,13 +3497,13 @@ mplate_has_keypoints_type: function (instance_template) {
        *
        */
       let keyframes_to_sequences = {};
-
+      console.log('sequence_list_local_copy', this.sequence_list_local_copy)
       for (let sequence of this.sequence_list_local_copy) {
         if (!sequence.keyframe_list) {
-          return;
+          continue;
         }
         if (!sequence.keyframe_list.frame_number_list) {
-          return;
+          continue;
         }
 
         let frame_number_list = sequence.keyframe_list.frame_number_list;
@@ -3906,18 +3906,15 @@ mplate_has_keypoints_type: function (instance_template) {
 
     },
     add_instance_to_frame_buffer: async function (instance, frame_number) {
-      console.log('add_instance_to_frame_buffer 000',)
       if (!this.video_mode) {
         return;
       }
-      console.log('add_instance_to_frame_buffer 111',)
       if (frame_number == undefined) {
         throw "frame number undefined in video mode (add_instance_to_frame_buffer)";
       }
       if (instance == undefined) {
         throw "instance is undefined in add_instance_to_frame_buffer()";
       }
-      console.log('add_instance_to_frame_buffer 111.5',)
       instance.creation_ref_id = uuidv4();
       instance.client_created_time = new Date().toISOString();
       if (this.instance_buffer_dict[frame_number]) {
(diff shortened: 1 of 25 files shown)


PJEstrada:
After some refactoring and more in depth debugging I found the way to replicate the error.

Actions To Replicate error:

1. Create a box
2. Copy it with ctrlc
3. Change frames with slide
4. Inmediately paste with ctlr+v while frame change is loading

Notes:

* I refactored the frame_number to be sent as a parameter on most places of annotation_core.
* 
* Save now has an extra check to save not only the current frame but also any frame that might unsaved according to `instance_buffer_metadata` and `unsaved_frames` list.

• After refactoring I was still getting the error, and noticed another issue. The problem here is with watcher/events. Whenever we change a frame, we are not blocking the saving if the instance_list of the current frame has not loaded. So if we paste quickly we will end up sending only 1 instance (the pasted instance) and all existing instances (the ones that had to load from frame buffer) will not be available and marked as soft deleted.

So to fix the above issue I refactored a couple of thigs:

1. I removed the get_instances() watcher. It's not used anymore.
2. `video.vue` component emmited an event called `change_frame_from_video_event`. Although it worked fine, we were not timing it correctly as we did not have a way to know when the instance fetch and frame loading in general was finished.
3. So I removed this event and splitted out into 2 `@go_to_keyframe_loading_started` and `@go_to_keyframe_loading_ended` now when we click the slider, we know there is a new frame being loaded and we can block in saves on there (triggered by pasting/ hotkeys). And when the frame is finished loading we trigger `go_to_keyframe_loading_ended` so that annotation core can load instances and do any other setup before unlocking the guard to allow saving.

anthony-sarkis:
Great going Pablo! :)

And good catch with the event case!  The division of the single frame change event into the start and end sounds good

frontend/src/components/annotation/annotation_core.vue
```diff
@@ -3886,45 +3921,42 @@ mplate_has_keypoints_type: function (instance_template) {
...
+    add_instance_to_file: async function(instance, frame_number = undefined){
+      if(this.video_mode){
+        if(frame_number == undefined){
+          console.error('Please provide a frame number to call add_instance_to_file()')
+          return
+        }
+        this.add_instance_to_frame_buffer(instance, frame_number)
+      }
+      else{
+        this.push_instance_to_image_file(instance, frame_number)
```
anthony-sarkis, line 3937: Can we get rid of `frame_number` entirely for this function? or is there a directional reason to keep it?

frontend/src/components/annotation/annotation_core.vue
```diff
@@ -3861,18 +3875,39 @@ mplate_has_keypoints_type: function (instance_template) {
...
+        this.instance_buffer_metadata[frame_number].pending_save = value;
+
+      } else {
+        this.instance_buffer_metadata[frame_number] = {
+          pending_save: value
+        }
+      }
+      // Keep unsaved_frames list to enable/disable save button
+      if(value){
+        this.unsaved_frames.push(frame_number)
```
anthony-sarkis, line 3898: This is excellent Pablo, it makes the video file abstraction stronger, a save saves the entire file across frames.

anthony-sarkis (approved):
Overall looks really good Pablo
It sounds like there may be some little adjustments in progress with tests etc but just putting the checkmark here since overall direction looks good to me

frontend/src/components/annotation/annotation_core.vue
PJEstrada: Thanks Anthony! Yes good catch :) I already removed it here: https://github.com/diffgram/diffgram/pull/491/commits/db6c15a034e003a9f63889cbf701105f4c4bc869

PJEstrada:
Great thanks for the review Anthony! I'm going to take a bit more time to add some tests to the newly touched functions so that we can start building a stronger unit test suite for annotation core.

There are also some tests I need to do with the rest of instance types as the CreateCommand refactor affects all of the types. I'll let you know when that's ready and if you find any issue while using it on video/image case please let me know :)

frontend/src/components/annotation/annotation_core.vue
anthony-sarkis: Thanks Pablo! :)

PR #497, opened 2022-01-12
Rehash Existing Instances to Prevent Instance History Corruption

PJEstrada:
## Context:
When the hashing algorithm was changed, we were not able to differentiate between a new hash from a user edit, or a new hash from the new algorithm version. 

Now we rehash the entire existing instance_list to always have the latest version of the hash in place.

- Added a test case to replicate this issue
- Pending some more unit tests to check instance history is not getting corrupted.

### DIFF
--- a/shared/annotation.py
+++ b/shared/annotation.py
@@ -657,7 +657,7 @@ class Annotation_Update():
             instance.hash_instance()
             new_hash = instance.hash
             if prev_hash != new_hash:
-                logger.warning('Warning: Hashing algorithm upgrade Instance ID: {} has changed \n from: {} \n to: {}'.format(
+                logger.info('Warning: Hashing algorithm upgrade Instance ID: {} has changed \n from: {} \n to: {}'.format(
                     instance.id,
                     prev_hash,
                     new_hash
--- a/shared/annotation.py
+++ b/shared/annotation.py
@@ -447,7 +447,7 @@ class Annotation_Update():
         return True
 
     def append_new_instance_list_hash(self, instance):
-
+        print('appending ', instance.id, instance.hash)
         if instance.soft_delete is False:
             self.new_instance_dict_hash[instance.hash] = instance
             return True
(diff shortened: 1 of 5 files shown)


shared/annotation.py
```diff
@@ -642,12 +643,30 @@ def detect_and_remove_collisions(self, instance_list):
...
+            new_hash = instance.hash
+            if prev_hash != new_hash:
+                logger.warning('Warning: Hashing algorithm upgrade Instance ID: {} has changed \n from: {} \n to: {}'.format(
+                    instance.id,
+                    prev_hash,
+                    new_hash
+                ))
+                self.system_upgrade_hash_changes.append([prev_hash, new_hash])
+            result.append(instance)
+            self.session.add(instance)
```
anthony-sarkis, line 667: Should we only add to session if the change as occurred?  Context that we use that session add control in other contexts later in the saving process / if there's no changes there's no need to add it yet?

shared/annotation.py
```diff
@@ -642,12 +643,30 @@ def detect_and_remove_collisions(self, instance_list):
...
         return result
 
+    def rehash_existing_instances(self, instance_list):
+        result = []
+        for instance in instance_list:
+            prev_hash = instance.hash
+            instance.hash_instance()
+            new_hash = instance.hash
+            if prev_hash != new_hash:
+                logger.warning('Warning: Hashing algorithm upgrade Instance ID: {} has changed \n from: {} \n to: {}'.format(
```
anthony-sarkis, line 660: Super small suggestion - I'm ok either way
I noticed we get some support tickets where people question the memory `warning` thing. If the default use case here is system upgrades, should we do `info` maybe?

anthony-sarkis:
Great work Pablo! I think your idea of doing the refresh on the new instances was really excellent here. And I like the implementation and the tests.  Good going!

shared/annotation.py
PJEstrada: Sure Anthony! That's a good idea I guess is less invasive for end users that way :)

shared/annotation.py
PJEstrada: Thanks Anthony! Added it here: https://github.com/diffgram/diffgram/pull/497/commits/ceccfd0d9250587c4e9e6854cc003dfb80594453

PJEstrada:
> Great work Pablo! I think your idea of doing the refresh on the existing instances was really excellent here. And I like the implementation and the tests. Good going!

Thanks Anthony! Already fixed the things you pointed out :) I'll wait for tests to pass and merge to master when ready

anthony-sarkis (approved):

anthony-sarkis:
Awesome!

anthony-sarkis:
Cool! :)

PR #507, opened 2022-01-14
Allow Batched Prelabels Upload With S3 Connectors

PJEstrada:
Batch object was missing the 

` batch.upload_aws_id = response['UploadId']`

for it to proceed with the batch upload of big prelabeled files.


anthony-sarkis:
Good catch Pablo! 😊

PR #532, opened 2022-01-25
Text interface v2

vitalii-bulyzhyn:


PJEstrada (commented):
Hey Vitalii!

I'm currently testing the interface. Good work on all of this! here are some of the bugs/notes I've found.

1. When we visit the url the only the file ID lile this for example: http://localhost:8085/file/57?frame=0&file=57&

image

We get the above error, where the instance list fetch fatils due to project_string_id not available on URL. We can fallback to getting the project_string_id from vuex store when this happens.


2. The `text_annotation_core.vue` looks pretty good. One idea to improve code readability (not necessarily for V1) is to wrap the svg elements into other components like `text_token.vue` `line_relation.vue`, etc.

3. There are a bunch of key errors on the console. I thinks these should be fixed before deploy to have a clean slate and be able to debug "real" issues once the code goes live:
image

4. When I try the relation creation I'm getting this error. This is not always but just for you to notice that.
image

5. I'm noticing that in order to create the relation, I have to be really precise with the mouse and  hover precisely over the line instead of either the line or the text. This makes it a bit dificult to create the relations. I'm not sure how hard that is to fix, so that the hover area includes both the text and the line of the label.

PJEstrada (approved):

PR #581, opened 2022-02-09
Configure Selinux label

### LINKED ISSUE #580: Postgres permission errors with local db on Linux (Centos)
Using diffgram 0.11.15 works flawless under windows when running python install.py and after entering all the required info.

On Linux Centos 7 the following error appears:
docker logs diffgram_db_1
chmod: changing permissions of '/var/lib/postgresql/data': Permission denied
chown: changing ownership of '/var/lib/postgresql/data': Permission denied

python install.py is run as root, anything I am missing here?

MathijsNL:
Without this option the container doesn't work under Linux distros with Selinux
https://docs.docker.com/storage/bind-mounts/#configure-the-selinux-label

Windows seems to be unaffected by adding the :Z option

This fixes #580

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -104,7 +104,7 @@ services:
       - POSTGRES_HOST_AUTH_METHOD=trust
       - POSTGRES_DB=diffgram
     volumes:
-      - ${POSTGRES_DATA_DIR:-./postgres-data}:/var/lib/postgresql/data
+      - ${POSTGRES_DATA_DIR:-./postgres-data}:/var/lib/postgresql/data:Z
     ports:
       - 5432:5432
 
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -104,7 +104,8 @@ services:
       - POSTGRES_HOST_AUTH_METHOD=trust
       - POSTGRES_DB=diffgram
     volumes:
-      - ${POSTGRES_DATA_DIR:-./postgres-data}:/var/lib/postgresql/data:Z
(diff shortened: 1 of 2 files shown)


anthony-sarkis:
Thanks @MathijsNL !  :)   Good solve! And thanks for testing both cases too.

I see the following text in the docs there:

> Use extreme caution with these options. Bind-mounting a system directory such as /home or /usr with the Z option renders your host machine inoperable and you may need to relabel the host machine files by hand.

I'm not clear if it's relevant to this case. 
If it is, we could make this extra `:Z` optional

@PJEstrada  if you could look at this too please I don't know much about this area

PJEstrada:
Hi @MathijsNL , thanks a lot for pointing this out and fixing this.

I was reading this: https://github.com/moby/moby/issues/30934

And some users mention that on linux it can potentially have dangerous behaviour due to the container having access to the host machine.

Although on our case most likely that won't be the case, I think we can add it as a non-default option during the install process and maybe ask users to add this extra label in case of using Selinux. 

I'm going to add this into the install.py script questions and leave the default as no Z flag. Then users that know what this flag means and it's implications can just change the default to "Add Z flag" and we would be able to better support selinux users.

Thanks again for the contribution!

MathijsNL:
@PJEstrada That is a valid concern. No need to force people to use a flag with dangerous behaviour.

Both of you thanks for the quick reply and the alternative fix.

PJEstrada:
Hi @MathijsNL,

I've added the new option to the `install.py` script. Can I trouble you to try it out to see if it fits your use case now?

MathijsNL:
@PJEstrada Yes perfect. This does exactly what it should now!

PR #591, opened 2022-02-12
Fix external database settings

### LINKED ISSUE #582: Using external postgres database fails under Linux
Hi there,

When using diffgram 0.11.15 with a remote postgres database everything works fine under windows.

Under Linux (Centos 7) there seems to be a problem with the database connection. During the python install.py the database connection seems to be successful, but in the diffgram_default_1 container it shows the following errors:

**docker logs diffgram_default_1**
psql: could not translate host name "db" to address: Name or service not known
Postgres is unavailable - sleeping
psql: could not translate host name "db" to address: Name or service not known
Postgres is unavailable - sleeping
psql: could not connect to server: No route to host
        Is the server running on host "db" (172.X.X.X) and accepting

I haven't found where the host name "db" gets defined so I can't fix it right now.  I also have no clue why it tries to connect to a 172.X.X.X IP as I haven't filled that in anywhere.

Hopefully someone can point me in the right direction.

MathijsNL:
This will fix #582

Most things discussed there are included I think, also the different variation of the connection url.

I tested it and it works for external db, didn't test it for internal db.

The db-init.sh might need some cleaning up, not sure if we need the parameters and I am unsure what the following lines are doing in there:

```
shift
cmd="$@"
```

$host parameter can be replaced by $DATABASE_HOST


PJEstrada:
Hello @MathijsNL ,

Thanks for the PR. I was testing it with the install.py script and it seems there is an error on the `docker-compose.yaml` syntax:

```
Enter diffgram version: [Or Press Enter to Get The Latest Version]: 
Do you want to use the local database? Y/N [Press Enter to use Local DB]: n
Please provide the database host: localhost
Please provide the database name: diffgram_e2e_tests
Please provide the database username: postgres
Please provide the database password: postgres
Testing DB Connection...
✓ DB connection succesful!
Do Add Z Flag to Postgres Mount? (Use only when using SELinux distros or similar) Y/N [Press Enter to Skip]: 
Do you want to add Mailgun to Diffgram?[Y/n] n
Generating Environment Variables file...
✓ Environment file written to: /home/pablo/Dropbox/diffgram/opencore/diffgram/.env
Launching Diffgram...
ERROR: yaml.parser.ParserError: while parsing a block mapping
  in "./docker-compose.yaml", line 1, column 1
expected <block end>, but found '<block mapping start>'
  in "./docker-compose.yaml", line 61, column 3
✓ Diffgram Successfully Launched!
View the Web UI at: http://localhost:8085
```

The rest of the code looks good to me :) Let me know when you fix the yaml thing and we can merge this to master.

MathijsNL:
@PJEstrada  It seems to be related to [9f6b0de](https://github.com/diffgram/diffgram/commit/9f6b0dea055926727ec8b79a6b7dbe13ee163150)

I think the spaces don't match up anymore.

MathijsNL:
@PJEstrada  If the section from **dispatcher** (line 3) to the line before **default** (line 60) get 1 space removed it should work.
Whiteline can be removed as well in between them.

I am not going to change and commit, because the diff shows up as a large removed and added section which is harder for you to check.

PJEstrada:
> @PJEstrada If the section from **dispatcher** (line 3) to the line before **default** (line 60) get 1 space removed it should work. Whiteline can be removed as well in between them.
> 
> I am not going to change and commit, because the diff shows up as a large removed and added section which is harder for you to check.

Thanks @MathijsNL, I've corrected the docker compose file and added some extra fixes for the local db case. Let's wait for tests and we can merge this to master

PR #593, opened 2022-02-12
Change file path from original file to the actual thumbnail file

### LINKED ISSUE #592: Thumbnail resolution is the same as the original image
Thumbnail resolution of images is the same as the original image. This results in an approximately equal or larger file size (especially for PNG images!).

Diffgram version 0.12.3 with Google Cloud Storage.

Edit:
File path pointed to (I think) the original file.

MathijsNL:
I think the original file got uploaded instead of the thumbnail. Result is shown below (file 14 + thumb is old code, file 15 + thumb is the new code)

image

This will fix #592

### DIFF
--- a/walrus/methods/input/process_media.py
+++ b/walrus/methods/input/process_media.py
@@ -1835,7 +1835,7 @@ class Process_Media():
         self.new_image.url_signed_thumb = url_signed_thumb
 
         data_tools.upload_to_cloud_storage(
-            temp_local_path = self.input.temp_dir_path_and_filename,
+            temp_local_path = new_temp_filename,
             blob_path = self.new_image.url_signed_thumb_blob_path,
             content_type = "image/jpg",
         )


MathijsNL:
I proposed another change in this pull request to use the JPG format when saving the file. This is (probably) not changing the originally uploaded file, only the resulting images that show up in the browser.

Personally I haven't seen too much difference between my PNG file and the resulting JPG that shows up in the browser but the decrease in size is significant.

Total storage for 1x PNG with original size of 1.4MB was 4.0MB (1.4 original file, 1.2 for the image/jpg and 1.4 for the "thumbnail")
With the proposed changes this will be a total of ≈ 1.8MB

I think the 95% JPG quality would be fine as a default setting, and if people prefer lossless quality there should maybe be an option for it in the settings? Or probably even a toggle to use the original uploaded images in the web interface?

PJEstrada (changes requested):
Hey @MathijsNL, thanks a lot for this PR. 

I think using the jpg for the thumbnails is a good idea. On the original image though, we might want to keep the same image quality and format, as some users are really detailed on their annotations and do require keeping the format and image quality (using zoom to annotate microscope bacterias pictures for example).

So I think if you could just keep the original image untouched we can merge this PR. 

Thanks again for the optimization, the spaced reduction is quite big for the thumbnails!

PJEstrada:
Hey @MathijsNL,

You're right about that, we're not actualy ussing the original image. The reason why we just rewrite for now and keep it separate is more a design decision for bein able to apply transformations during upload while keeping the original image. This is not implemented yet but it's more a future direction choice. For example, the user might want to add a greyscale or increase brightness to their image. For this, we would keep the original image on one blob, and the "transformed" image (which currently is not actually transformed) on another one.

What I'm not sure though, is if just doing a rewrite of the image with `im.write` is actually altering something on the original image. Up until now we haven't had any complains on it but maybe there's some insights you can give us on that?

MathijsNL:
Lossless image formats (PNG is the only one I tested) don't change as far as I can see.
Lossy image formats like jpeg do!

I saved the striped PNG as quality 100 JPG.
Afterwards I saved the quality 100 JPG as quality 75.


image

I guess it would be safe to save lossless formats, but for lossy formats the original file maybe should be used?

@PJEstrada Let's discuss the saving of files in another topic, this is getting out of the scope for this PR

MathijsNL:
@PJEstrada This PR only contains the thumbnail change. Can you review and merge?

PJEstrada (approved):

PJEstrada:
Thanks for the PR @MathijsNL!

Also thanks for the test there on the formats. It seems the right route is just what you mentioned (Adding compression only on lossless). Feel free to add a new pr or a new issue if you feel you still might need to discuss this a bit.

anthony-sarkis:
Hey @MathijsNL thank you very much for the update and the enlightened discussion here! Truly much appreciated!! :)

I'll open new item to help continue the discussion 
https://github.com/diffgram/diffgram/issues/596

PR #606, opened 2022-02-16
Fix image compression on facing media

### LINKED ISSUE #596: Image compression on using facing media
Continuing the discussion from https://github.com/diffgram/diffgram/pull/593

To summarize some of the above:
1) We currently store the raw files as part of ingestion process. So the the assumption when creating any new modified artifacts is they are for specific purpose. For example for human UI centric use. The user or other processes can decide to pull the raw or modified image.
2) In the context of 1) we may as well focus on space efficient storage options, e.g. for network transfer speed to user etc. For example defaulting to JPG and even a specific value on imwrite like 95

To add another reason why we do the raw storage, is that it helps decouple the processing layer from the raw ingestion layer. So for example as long as we can stream the raw bytes to the server we can ingest as much as we want. And if we have say a 30GB video file that needs to retry processing we can do that too.

This discussion is generally focused on images, but can also apply to other media formats.
Video -> We assume the new artifact will be .mp4 

Current state (to be verified)
Assuming this is already happening with imwrite default settings if input is jpg already
Assuming this is not happening if the input type is `png`

MathijsNL:
This fixes #596

New behavior for lossy formats is to save original raw image as facing media.

New behavior for PNG is to only rewrite the file if compress_level=0 or 1. It will apply compress level 2 in that case.

Lossy formats can be verified by uploading jpg through diffgram and download from cloud storage and compare the hash.
PNG format can be verified by uploading any PNG with compress_level=2..9 and download from cloud storage and compare the hash. For level 0 and 1 you can upload and see the file size decrease (although the difference between level 1 and 2 is small sometimes).

As for the reason I didn't go for the idea of using the filesize is because there isn't a reliable way to determine the alpha channel after it is removed.

The check for compression level takes about 0.7 seconds per 10.000 PNG files (reading the byte and assigning the Boolean). 

Altering this to use default compression is possible as well. I leave that design choice to you.

Just need to check for Superfast and Fast compression
```
compress = ( f.read(1) in [b'\x01', b'^'] )
```

And change the compress_level to 6
```
imwrite(new_temp_filename, np.asarray(self.raw_numpy_image), compress_level=6)
```

Bulk uploads of 10K images could take up to an hour to process if the compression needs to be applied to all of them in case level 6 gets chosen. The space saved for images at level 5 could be 470mB, and for to level 2 and 6 could be 4.340mB.


PJEstrada (approved):
Thanks for the optimization here! Great research on these topics. I think your proposal for compression on only png files is good in my opinion :)

I just merged master to this and will wait when tests pass to close PR

Thanks again for all your contributions @MathijsNL

PR #611, opened 2022-02-17
fix: change panning to shift key

PJEstrada:
Now we pan using the shift key on image/video UI


anthony-sarkis:
Great! Thanks Pablo

anthony-sarkis (approved):

PR #638, opened 2022-02-19
Add support for BMP, TIF and TIFF files

### LINKED ISSUE #240: [backlog] Discussion on Lossless formats like TIF/TIFF
Workaround example:
https://stackoverflow.com/questions/28870504/converting-tiff-to-jpeg-in-python  

Main contexts here is
1) What do we need to do reasonably render it on frontend
2) Storing the original, in the context of Diffgram being the one database for all of the data.

**I'd like to get more peoples insights into the need for this and desires here.**

one context, material science:
"microscopes provide .tif, which we like as it typically stores the acquisition metadata and pixel information without quality loss."

MathijsNL:
This PR will close #240  I think.

.bmp, .tif and .tiff files will be supported. Original file will go to raw files and a PNG with compress_level=4 is generated because it feels like a good tradeoff between the required processing power and the space saved. Feel free to adjust this as needed. 

Because thumbnail would be generated with the native file extension this approach didn't work. All thumbnails will be generated as JPG with quality=95.

If a new extension gets requested it can be added to the same list as .bmp, .tif and .tiff as long as imageio can read and write it the same way as the other extensions.

@PJEstrada I changed the thumbnail extension to JPG because it is a thumbnail and you don't really see the difference with PNG. It is slightly smaller and blazing fast. It is in a separate commit, so you can also get the thumbnails as .PNG.

Here is a small comparison between a JPG and PNG thumbnail to visualize it
image

### DIFF
--- a/.github/workflows/diffgram_testing.yaml
+++ b/.github/workflows/diffgram_testing.yaml
@@ -6,7 +6,7 @@ on:
       - 'releases/**'
       - '**'
   pull_request_target:
-    types: [review_requested]
+    types: [review_requested, synchronize, ready_for_review, opened]
     branches:
       - '**'
       - '!DAD**'
--- a/.github/workflows/diffgram_testing.yaml
+++ b/.github/workflows/diffgram_testing.yaml
@@ -8,7 +8,7 @@ on:
   pull_request_target:
     types: [review_requested]
     branches:
-      - 'releases/**'
+      - '**'
       - '!DAD**'
 jobs:
   Default-Service-Unit-Tests:
--- a/frontend/src/components/input/new_or_update_upload_screen.vue
+++ b/frontend/src/components/input/new_or_update_upload_screen.vue
@@ -291,7 +291,7 @@
           loading_annotations: false,
           upload_source: null,
           accepted_annotation_file_types: ['json', 'csv'],
-          accepted_files: ".jpg, .jpeg, .png, .mp4, .m4v, .mov, .avi, .csv, .txt, .json",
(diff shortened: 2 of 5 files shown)


anthony-sarkis:
Good going!! Thank you very much for adding this extended support.

I read the description and skimmed the code. It looks good to me!

Our usual schedule it Monday to Friday - so Pablo will be back on Monday and can give his review / final ok.

PJEstrada (approved):
Hey @MathijsNL,

Thans for the new support on these files! I checked the code and it looks really good.

I just added a commit with some unit tests for it. We are now trying to increase our testing coverage so I added this as a sample for you to see how we are adding unit tests to new code we add.

Here is a bit more info on that: https://diffgram.readme.io/docs/unit-testing-backend-guidelines

I will wait for tests to pass and merge once ready :)

Thanks again

PJEstrada (approved):

MathijsNL:
It was nice to add support for these files, just in case you have some in your dataset (I have some datasets that are 50% BMP and too lazy to convert them).

Could you point out where the unit tests get their images from?

I will try to add a unit test as well on my next contribution.

PJEstrada:
> 

Hi @MathijsNL, 

Since we are only doing unit tests for now. We try to mock the data to follow the key conditions that we check on the file. So the easiest way here is to create a temp file and write the byte you are checking like this:
```

        temp = tempfile.NamedTemporaryFile(suffix = '.png')
        with open(temp.name, 'wb') as f:
            f.seek(63)
            f.write(b'\x01')
```

We can then mock the call `imwrite` to simulate the writing so we don't tests the specifics of the `imwrite` result (that should be testsed on the imageio library)

You can look at `test_process_media.py` to see all the details on how we setup the image and mock external library calls.

MathijsNL:
Thanks for the explanation!

PR #639, opened 2022-02-19
Make Issues Concept more Visible & Fix global attribute label file permission

anthony-sarkis:
### Fixes
1) Fix global attribute label file permission https://github.com/diffgram/diffgram/commit/05b7d1f09dfb1d3318c4f2f6ff049e6f71f29cd7
2) Fix undefined https://github.com/diffgram/diffgram/commit/37b756379b683e8bf5abb886543d4b66592b0f64
3) Fix bug with Scripts expansion panel https://github.com/diffgram/diffgram/commit/fef0fdf12697170dfa9e5fe8618c0c210abe2686

### Make issues more visible by default 
rest of commits starting with -> 
https://github.com/diffgram/diffgram/commit/35c8e8b614c1e05fccd955e9d6ee0f8e9d745117

To recap highlights of that:
1) Make issues according viewable by default. Before had to create an issue to see the existing issues (in this context).
2) Add a "new issue" button. This means a user does not have to know to right click on canvas. Also makes this accessible to other media types more easily
3) Add a "no issues" style art to visual. Fix v-cloack

#### Now visible as an expansion panel:
Key point -> it wasn't visible here at all before. To even get the left to come up, had to click on canvas and know about feature. 
image

#### Before and After
UI/UX wise now this should make it easier for the user to dive into describing the issue.
image

### Notes
1) Going through this discovered some various unexpected behavior. As far as I can tell it's not directly from this update, but just more visible now that the components themselves are more visible. (for example the `issues_list` does not seem to update very well.)

### DIFF
--- a/frontend/src/components/annotation/annotation_core.vue
+++ b/frontend/src/components/annotation/annotation_core.vue
@@ -358,6 +358,7 @@
             <v-expansion-panel-content>
 
               <tooltip_button
+                  v-if="show_modify_an_issue != true"
                   tooltip_message="New Issue"
                   datacy="new_issue_in_side_panel"
                   @click="show_modify_an_issue=true"
--- a/frontend/src/components/annotation/annotation_core.vue
+++ b/frontend/src/components/annotation/annotation_core.vue
@@ -274,42 +274,7 @@
             :instance="selected_instance_for_history"
           >
           </instance_history_sidepanel>
-          <create_issue_panel
-            :project_string_id="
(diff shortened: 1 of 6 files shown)


PJEstrada (approved):
Hey Anthony! Thanks a lot for adding this :) I just checked the code and it looks good to me. I'll merge to master now

anthony-sarkis:
Thanks Pablo! :)

PR #669, opened 2022-02-24
Fstring conversion, first iteration

### LINKED ISSUE #632: Python coding style choices
There are a lot of different coding styles used within Diffgram.

- String concatenation
- .format strings

Python f-strings are not used at all except for one instance in the install.py script.

In my opinion f-strings are more readable. For people that don't know enough about this topic I will illustrate this with a simple example.

```python
# String with .format
subject = "[{}]{} Has Shared an Instance With You ".format(self.project_string_id, self.user_sending.first_name)

# F-String

subject = f"[{self.project_string_id}]{self.user_sending.first_name} Has Shared an Instance With You "

```

@PJEstrada if you agree I can do the conversion. Let me know

MathijsNL:
This PR will take care of almost all the old .format strings and string concats, I didn't count but if we leave the @routes.route() functions out there is at most 10 files with a few exceptions that need to be looked at so we probably got 95% or more with this change.

If this PR gets merged it will fix #632 I think. It might be a good idea to create a new issue then for the remaining files and the fixes for str() conversions within fstrings.

# Converted files
Most of the code files, test files and the installer.
Most fstrings have double quotes except for the few that use double quotes inside.

# Partially conversion
Some fstrings have a str() variable inside after conversion. Example:
```
kind = f"api_enable_{str(input['api_name'])}",
```
It probably won't hurt, and it can most likely be removed. Saving for next iteration.

# Not converted
All instances of @routes.route(). Updating them didn't really improve readability. These files should probably stay as they are, unless someone can think of a way to make these more readable.

```
# Old
@routes.route('/api/v1/project/<string:project_string_id>' +
              '/userscript/list',
              methods=['POST'])
# New
@routes.route(f"/api/v1/project/<string:project_string_id>/userscript/list",
              methods=['POST'])

```

Other files:

- default\methods\source_control\file\file_browser.py
- local_dispatcher\env_adapter.py
- shared\settings\env_adapter.py
- shared\helpers\performance.py

Reason:
No straightforward conversion

@PJEstrada I don't have time to build my docker containers and test right now so I am just creating the pull request.
I separated the commits to the code and test files so you could more easily run your tests without modifications if that would be needed.

Also do you agree on double quoted fstrings by default?

### DIFF
--- a/default/methods/share/share.py
+++ b/default/methods/share/share.py
@@ -380,16 +380,16 @@ class Share_Project():
         if self.notify is False:
             return
 
-        subject = "Added to project " + self.project_string_id
+        subject = f"Added to project {self.project_string_id}"
 
-        message = " You have been added to: " + self.project_string_id
-        message += " as a " + str(permission_type)
+        message = f" You have been added to: {self.project_string_id}"
+        message += f" as a {str(permission_type)}"
 
         message += ". Access the project here. " + settings.URL_BASE + \
                    "project/" + self.project_string_id
 
-        message += " Added by " + str(self.user_who_made_request.email)
-        message += " With personal note of:" + str(note)
+        message += f" Added by {str(self.user_who_made_request.email)}"
+        message += f" With personal note of:{str(note)}"
 
         communicate_via_email.send(self.user_to_modify.email, subject, message)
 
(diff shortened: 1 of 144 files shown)


anthony-sarkis:
Great going!! This will make it much more readable and consistent! Thanks @MathijsNL

PJEstrada (commented):
Hey @MathijsNL ,

Thanks a lot for this refactor! I'm going to be checking each file shortly and adding any comments on missing parts or any other thing I find. Thanks again for this big PR :)

default/methods/share/share.py
```diff
@@ -380,16 +380,16 @@ def email_existing_user(
...
 
-        subject = "Added to project " + self.project_string_id
+        subject = f"Added to project {self.project_string_id}"
 
-        message = " You have been added to: " + self.project_string_id
-        message += " as a " + str(permission_type)
+        message = f" You have been added to: {self.project_string_id}"
+        message += f" as a {str(permission_type)}"
 
         message += ". Access the project here. " + settings.URL_BASE + \
```
PJEstrada, line 388: I think this one is missing :)

MathijsNL: I used https://github.com/ikamensh/flynt to help with the conversion. Afterwards it was a lot of manual checking of all the changes to make sure the conversion was good. 

I just ran it again on this specific file, and for some reason it doesn't want to convert the one on line 388 there.

Good catch though!

PJEstrada (approved):
Thanks for all this work @MathijsNL,

I just found acouple strings missing but everything else seems good!. I'm approving this and waiting for tests to pass

MathijsNL:
You're welcome.

I will try to update the missing things once this PR has been completed.

default/methods/share/share.py
PJEstrada: Thanks @MathijsNL

PR #674, opened 2022-02-24
fix: target reticle display on template creation

PJEstrada:
Target reticle was disappearing when deleting a node while drawing a relation. This behaviour is now fixed.


PJEstrada:
This also fixes the issue of the entire keypoint instance dissappearing when resizing it

PR #679, opened 2022-02-25
Rescale Keypoints Snapping to Corners

PJEstrada:
Now we can draw and edit the keypoints instances rescaling based on fixed point.

Logic is as follow:
-  We scale fixing the point on the opposite direction of the control point.
- Example: we rescale on the right control point => the left control point stays fixed.
- We rescale the top right point => the bottom left point is fixed.


anthony-sarkis:
Great going Pablo this is going to be more intuitive for the user!!

anthony-sarkis:
Great going Pablo 😊🎉 💯

PJEstrada:
Thanks Anthony!!

PR #691, opened 2022-03-01
Show/Hide Instance Templates on UI Schema

PJEstrada:
We can now customize the schema to show hide instance template from the project.


anthony-sarkis:
Super Pablo! :)

PR #725, opened 2022-03-09
Add Guided Drawing Mode for Keypoints Instances

PJEstrada:
Now users can draw keypoints one point at a time and draw them in a customizable order.


anthony-sarkis:
Hey Pablo! Great work!! **Amazing feature!!!**

I made two small little suggestions
[Suggestion, header 1 for primary message](https://github.com/diffgram/diffgram/pull/725/commits/f8ef045df8e6bb66d1716e1aa488d84d61591db2)
image

[Improve: Scale points typo](https://github.com/diffgram/diffgram/pull/725/commits/737d17bfde5060796f22b16b5e10d79ec2f4c2bd)

After that, I noticed the control points don't always seem to line up with the box
Perhaps it was dependent on the specific size of the thing prior and also needs to scale or something
image

One small thing I noticed in QA was that the occlusion effect now seems to be only visible on hover (I believe the expected state is that occlusion effect visible without hover)
image

PR #727, opened 2022-03-10
Improve feature checker

anthony-sarkis:
[Improve/Fix: Where the plan comes from](https://github.com/diffgram/diffgram/commit/af7452f7126d8b23ffb2254476da799efe207646)
[Improve: Move super admin bypass to generic spot](https://github.com/diffgram/diffgram/commit/07996ae40a88b1de8ec2aba66a0011d2c5015468)

@PJEstrada Please check and what do you think?

*Edit. The failing tests seem kind of odd not sure how it would be related to these changes. Maybe the test runner is interacting oddly with the plan thing somehow?
image

### DIFF
--- a/default/methods/share/share.py
+++ b/default/methods/share/share.py
@@ -182,10 +182,6 @@ class Share_Project():
 
     def check_free_tier_member_limits(self):
 
-        if self.user_who_made_request.is_super_admin is True:
-            if settings.IS_OPEN_SOURCE  == False:
-                return True
-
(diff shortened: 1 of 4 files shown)


PJEstrada (approved):
Thanks Anthony! I added a couple commits to fix the errors on e2e and remove export limit on list query.

anthony-sarkis:
Ohhh good catch!! I forgot plan could be None in that case! thanks Pablo!!

PR #731, opened 2022-03-10
fix: Fix billing issue for users with no accounts.

PJEstrada:
This can happen if a new user gets invited to a project, or if a user tries to upgrade and has no accounts yet.

We are adding a fallback method to the credit card payment so that if the user does not provide an account, it will use the first one on the account list and if it does not have any accounts. It will create a new billing account for the user.

This a bug on the order_premium component where the account_info component was not getting properly imported and thus, it was not fetching the account data.


anthony-sarkis (approved):
Excellent work Pablo!! :)

PR #736, opened 2022-03-11
DAD-132 Adjust Rescale Control Points for Instance Templates

PJEstrada:
Now scale control points have better scales and a bit more "invisible drag space"


anthony-sarkis:
Awesome!! :)

PR #753, opened 2022-03-18
Dad 163 add tree structure attribute type frontend view

vitalii-bulyzhyn:
In this branch:

- adding UI for the creating tree attributes
- add tree component to annotation UI
- wire up new tree attribute to the backend


PJEstrada (changes requested):
Hey Vitalii! Thanks for all this work on this :) Code is looking good overall

I think the PR is still WIP but There are a few items I got while testing this out and reading the code. Maybe you are already aware of some of this, but just sharing fyi

1. I think the logic for rendering an existing tree is missing on the wizard. This is in the context of editing and already existing attribute. To replicate, create a new tree view attribute, add some items and refresh page. You wont see those items after refreshing/
2. The rendering of the treeview is also missing on the annotation UI. 
3. I think some of the text UI and command pattern stuff got in this PR. I'm ignoring it for now since I think that's on a different context.

PR #762, opened 2022-03-29
Fix/tree building bug

vitalii-bulyzhyn:


frontend/src/helpers/tree_view/construct_tree.ts
```diff
@@ -42,6 +42,47 @@ export const build_path = (array: Array<any>, parent_id, path) => {
 }
 
 export const construct_tree = (node_list: Array<Node>): Array<any> => {
+<<<<<<< HEAD
```
anthony-sarkis, line 45: @vitalii-bulyzhyn this has a head tag  / document has has source control artifacts

vitalii-bulyzhyn: Just removed

PJEstrada (changes requested):
Hey Vitalii, I just did more testing here and found that the attribute searchbox is displaying on other attribute types too:

image

I think once fixed we can merge, did not found anuthing else up until now

anthony-sarkis:
Thanks everyone

PR #767, opened 2022-03-29
Tiny fix settings

anthony-sarkis:
[Add: View existing members from Share concept](https://github.com/diffgram/diffgram/commit/0157046fbd0ee59cc9afb40299d683d53ea9d3ec)

-> Add dialog to show existing users when inviting
-> Add extra button to jump back to user mode when in API creation (design intuition wise the button is in same place so the user doesn't have to think about the selector, otherwise if they view existing first then it feels confusing)
-> Upgrade and clean other buttons
-> Add flag to optionally not show the "existing users" e.g. on settings page
-> Clean username / move client ID to otherwise empty step
Thinking design perspective here, member_id maybe useful for debugging but not average user
-> Remove titles so parent component can set relevant title, or can completely remove it in cases where it's contextually obvious and a title is design clutter

Other small things:

- [Change on save to remove button](https://github.com/diffgram/diffgram/commit/2f9e8d59af3791f3f9ef8cb66e5ba176da43e796)[](https://github.com/anthony-sarkis) 
- [Reorder Members to be higher priority](https://github.com/diffgram/diffgram/commit/950b6b5c766af95be909bc7f376d43af73edd8af)[](https://github.com/anthony-sarkis) 
- [Hide user type and clean](https://github.com/diffgram/diffgram/commit/9b629f93c35c378e06d8ecdc270fa4406fb0a860)[](https://github.com/anthony-sarkis) 
- [Add contextual help button](https://github.com/diffgram/diffgram/commit/382ce9e42b57e64c26e7a0e13ea6fa3642254276)[](https://github.com/anthony-sarkis)

share project
show existing
dev auth
new settings clean

### DIFF
--- a/frontend/src/components/project/settings.vue
+++ b/frontend/src/components/project/settings.vue
@@ -57,24 +57,6 @@
           </v-card-text>
         </v-card>
       </v-container>
-      <v-container>
-        <v-card>
-          <v-container>
-
-            <h2> Resources </h2>
-
-              <v-btn color="blue darken-1"
-                      text
-                      @click="$router.push('/project/' + project_string_id + '/attributes')"
-                     >
-                <v-icon left>mdi-collage</v-icon>
-                Attributes
-              </v-btn>
-
-
-          </v-container>
-        </v-card>
-      </v-container>
 
 
       <v-container>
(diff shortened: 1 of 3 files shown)


PJEstrada (approved):
Thanks Anthony! All of it looks good to me :) If you're ok with it we can merge to master :)

Issue #775, opened 2022-04-01
Improve error handling on upload wizard when walrus service is down


PJEstrada:
Sometimes users get unexpected errors from misconfigurations on the walrus. But this errors are not correctly displayed on the UI. 

We need to add better error handling on the upload step and progress step of the upload wizard.

Screenshot from user on public support slackchannel:

![image](https://user-images.githubusercontent.com/6606958/161351301-b64891ff-98e5-4e93-83c7-b072a68213e2.png)

anthony-sarkis:
Also applies to cases where configuration is correct, but the container is down / not running
and especially relevant if the default service is up but the walrus is down

PR #777, opened 2022-04-02
Status checking & 503 fix

anthony-sarkis:
1. [Fix: 503 Service Unavailable](https://github.com/diffgram/diffgram/commit/d1e4bf79cb8defa9937afeb01975d74de8809144)
This will helps differentiate from issues within each service, so we know when it's a 503 service not available vs 500 internal error
2. Add status page. Status page will check services are up
After a little bit of reflection it seemed better to keep this seperate from install info.
3. Other little items like 
[Properly surface 404](https://github.com/diffgram/diffgram/commit/2c4d6cd080282fe32fd79ed0670fc4d9b52eb4f5)

Closes #776 

Status page is `/status`. And a link at the bottom of settings menu is available.

image

example
both ok
status  example

PR #778, opened 2022-04-02
Fix import success

anthony-sarkis:
_This builds on #777 so probably best to review that first_

### Automatic status checking
Now, it will automatically check if Walrus is up before the user even starts to work. 
This is mainly in the context of first time installs.

automatic service down detection

### Declare Success is better
Before it was assuming success if not actively uploading. In this V2 iteration we improve it to seperate the success concern from the progress concern.
declare success event

### Expressly render errors 
In fetch data and dropzone cases
Example, walrus is killed halfway through upload
check connection to walrus
dropzone case


anthony-sarkis:
closes #775 

@PJEstrada  You know more about this section of the system so please check if it covered the cases / didn't break anything please :)

frontend/src/components/input/new_or_update_upload_screen.vue
```diff
@@ -144,7 +144,9 @@
                               @vdropzone-removed-file="file_removed"
                               v-on:vdropzone-thumbnail="thumbnail"
                               @vdropzone-sending="drop_zone_sending_event"
-                              @vdropzone-complete="drop_zone_complete">
+                              @vdropzone-complete="drop_zone_complete"
+                              @vdropzone-success="drop_zone_success"
+                              @vdropzone-error="drop_zone_error">
```
PJEstrada, line 149: Nice! I did not know about this event handler

install.py
```diff
@@ -377,22 +377,22 @@ def populate_env(self):
             env_file = f"GCP_SERVICE_ACCOUNT_FILE_PATH={self.gcp_credentials_path}\n"
             env_file += f"CLOUD_STORAGE_BUCKET={self.bucket_name}\n"
             env_file += f"ML__CLOUD_STORAGE_BUCKET={self.bucket_name}\n"
-            env_file += 'SAME_HOST=False\n'.format(self.bucket_name)
```
PJEstrada, line 380: thanks for fixing this

PJEstrada (approved):
Thanks Anthony! Everything looks good to me, that dropzone handler is going to be super usefult for catching uploads in progress.

PR #779, opened 2022-04-02
Better query access, better discovery UI/UX

anthony-sarkis:
### Query from anywhere

The idea is to extend the usefulness of the data explorer by adding easy contextual buttons.
For example if viewing the Schema and I want to know what a label looks like , being able to jump to that query:

explore screen
example page

These functions already existed so it's as easy as adding a button.
- [] A follow up in different thing is to run our existing per label report structure so we get counts here as a preview.

### Surface Query and Discover Page in main menu better
Small amount of code more product angle
image
[Make Query Concept more clear](https://github.com/diffgram/diffgram/commit/69bff5bc03f5939b6b38f5a98c62596246d6cb31)

### Fix soft delete 
[Fix: Ignore soft delete by default](https://github.com/diffgram/diffgram/commit/1b380d26daf12239818d55a2f175a39fe99b3258)
this little one has been sitting around for a long time!!

### Various little UI improvements
1) Smoother suggestions [Improve ease of use](https://github.com/diffgram/diffgram/commit/71f515b3759a996977de4352acd9e69417ccb86b)
2) UI visible not found screen
3) Clear existing files
no results screen

Overall not a ton of code but mostly product and UI/UX work.

### New design proposal
Main ideas here
* More focus on the query as most important part
so larger font, more prominent etc
* one "task" at a time, so for example the compare models has a toggle
* left side bar seems to work better for this "text heavy" context
* left side bar seems to work better when restore down halfway, looks pretty sharp 
* 'project' should not be needed since always in the scope of one project at the moment

visual
build focus
new design example
compare enabled

### issues
[ ] linear loader in new design is not quite displaying right
[ ] the "no more results" is a bit awkward in some cases, eg infinite load etc
[ ] the difference between clicking away, execute, and refresh etc to run a query needs some more work


PJEstrada (changes requested):
Hey Anthony thanks for this! The new layout looks super good I think we have a lot more screen space to add more comples filters now.

One small thing I found though is a small layour issue with the empty placeholder:

image


You can replicate it by going fullscreen and doing a query that returns 0 results. Other from that all looks good to me!

anthony-sarkis:
Awesome!!

Thanks I will review

PR #785, opened 2022-04-04
Remove blipping when changing video frame

PJEstrada:
Removed dark bg when changing frames on video.

### DIFF
--- a/frontend/src/components/annotation/annotation_core.vue
+++ b/frontend/src/components/annotation/annotation_core.vue
@@ -7358,10 +7358,11 @@ export default Vue.extend({
          * seperetly for special event handling
          *
          */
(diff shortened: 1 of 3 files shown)


frontend/src/components/annotation/annotation_core.vue
```diff
@@ -7762,6 +7763,7 @@ export default Vue.extend({
           const new_image = await this.addImageProcess(file.image.url_signed);
           this.html_image = new_image;
           this.refresh = Date.now();
+          await this.get_parent_instance_list_for_video();
```
anthony-sarkis, line 7766: This was causing latency in regular use
Right now global is global to video not frame so here is ok for now
if we in future support per frame globals then would revisit
(from discussion with Pablo)

PR #802, opened 2022-04-07
Fix: Missing param

### LINKED ISSUE #799: Going to /explorer on studio does not appear to auto change project correctly
e.g. in Project A 
then go to https://diffgram.com/studio/annotate/coco-dataset/explorer
expected that it will auto change project

anthony-sarkis:
When we manually set props like this then it appears we need to also set default props. becuase it was not receiving a project string id the automatic project changer could not work

closes #799


anthony-sarkis:
Given how small the change is I will merge it

PR #820, opened 2022-04-19
Login new assumptions

### LINKED ISSUE #817: Old account being treated as a new account and menu icons greyed out
When I logged in to my Diffgram account (on Diffgram.com) using my ID and password, I was taken to the onboarding screen where my account (Active since 3+ months) has been treated as a new account.

None of my annotation progress was visible nor my labels, scheme or projects and as a user, was prompted to create new. 

**When I manually entered the URL suffix /me**, I was redirected to the /me page and I clicked on the annotate next task page which restored everything. 

Attaching a Loom video. 

https://www.loom.com/share/18d075bf14b245f9b9b9adc8cbaa6029

anthony-sarkis:
[Add dedicated change project button for rare contexts](https://github.com/diffgram/diffgram/commit/abc75715b9ac800c23bfd29c9b1aa06cb0432668) 

Keeping in mind the context that we normally set the project when the user logins in, in the rare event that there is no project:
image

1) The part that was confusing was that change project was within the project menu.
This meant that various conditions that made sense for most of the menu did not make sense for change project.

2) Now there is a dedicated change project, that will only show up when there is no active project, the user's email is verified, and there is a project available
3) Now the project and task menus only shows when there's an active project.
While there's some UI trade offs for now that seems simpler then the disabled

while we could "get smart" about some of the other resume onboarding, change, new project etc. there are so many contextual things for now feels better to keep that screen open - again keeping in mind it's a fallback that users should rarely get to

this iteration should close #817

### DIFF
--- a/frontend/cypress/integration/diffgram/user/user_login_spec.js
+++ b/frontend/cypress/integration/diffgram/user/user_login_spec.js
@@ -15,7 +15,6 @@ describe('Login Flow tests', function () {
       'ai',
       'alert',
       'public_project',
-      'annotation_assignment',
       'annotation_project',
       'annotation_state',
       'attribute',
--- a/frontend/cypress/integration/diffgram/user/user_login_spec.js
+++ b/frontend/cypress/integration/diffgram/user/user_login_spec.js
@@ -1,6 +1,5 @@
 
 
-
 describe('Login Flow tests', function () {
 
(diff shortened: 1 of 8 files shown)


anthony-sarkis:
This is a pretty small one so am planning to merge once tests pass

PR #829, opened 2022-04-22
Audio interface V1

vitalii-bulyzhyn:
- Annotate Audio Regions
- Audios Transcription (Global Attribute with Long Text Input)
- Waveform Navigation (Pan, Zoom In, Zoom Out)
- Attributes Support (Both Instance level and Global)
- Probably displayed below instead of left side
- Tab for instance, tab for global file?
- Timeline Visualization
- Task Mode
- Basic Hotkeys


PJEstrada (approved):
Awesome work Vitalii thanks!

PR #839, opened 2022-05-03
Feature checker logging

anthony-sarkis:
## This makes the feature checker log more visible
The context was debugging an issue, the issue turned out to be a combination of old assumptions mixed with new code (assumptions on who the user was), and env issues, etc. 
All would have been easier with more clear http logs so here it is.

## Future
A more general question is if we want it to be a standard practice to do this http instead of, or in addition to, regular logging.
That's part of a broader discussion. This is just one example where the regular logging wasn't that helpful but http log, getting the "why" at the same time as the result, is more help

## Examples

example log

now feature checker includes the log


PJEstrada (approved):
Super thanks Anthony! Everything looks good to me :)

anthony-sarkis:
Thanks Pablo! :)

PR #843, opened 2022-05-04
Text interface v1.2

vitalii-bulyzhyn:
- Add attributes to the text interface
- Modify tokenizer to handle paragraphs
- Render text with paragraphs on the frontend
- improve text annotation experience


PJEstrada (approved):
Thanks Vitalii!

PR #855, opened 2022-05-12
Clean old code

anthony-sarkis:
This was implemented fully in: https://github.com/diffgram/diffgram/tree/master/frontend/src/components/text_annotation


PJEstrada (approved):
Thanks Anthony!

PR #857, opened 2022-05-14
add minio support

ppolxda:
Use AWS S3 credentials. Support Minio, some problems encountered in commit

Minio Only Allow Expiry time is less than 7 days (value in seconds). 
[https://github.com/minio/minio/blob/ca69e54cb6e4cff32b39cef6c7231c6d7ee6fca6/cmd/signature-v4-parser.go#L232](https://github.com/minio/minio/blob/ca69e54cb6e4cff32b39cef6c7231c6d7ee6fca6/cmd/signature-v4-parser.go#L232)

I try to fix it in this commit.  But I don't think it's a good fix
[https://github.com/diffgram/diffgram/tree/0cff270151fa50f42f0fc580c5dbd85dacf0d14d](https://github.com/diffgram/diffgram/tree/0cff270151fa50f42f0fc580c5dbd85dacf0d14d)


anthony-sarkis:
@ppolxda Thank you very much for putting this together! :)

We will be back on Monday PST and review it in depth. 

Semi-related I just added https://github.com/diffgram/diffgram/pull/858
and that also may help with refactoring expiry

ppolxda:
> @ppolxda Thank you very much for putting this together! :)
> 
> We will be back on Monday PST and review it in depth.
> 
> Semi-related I just added #858 and that also may help with refactoring expiry

@anthony-sarkis 

I think determine_if_should_regenerate_url  function is gread.  and same problems in build_secure_url.

shared/data_tools_core_minio.py
```diff
@@ -0,0 +1,40 @@
+# OPEN CORE - ADD
+import boto3
+from shared.settings import settings
+
+from .data_tools_core_s3 import Config
+from .data_tools_core_s3 import DataToolsS3
+
+
+class DataToolsMinio(DataToolsS3):
```
PJEstrada, line 9: Inheritance here was a great idea!

shared/settings/settings.py
```diff
@@ -174,6 +181,8 @@
 print("PROCESS_MEDIA_REMOTE_QUEUE_ON", PROCESS_MEDIA_REMOTE_QUEUE_ON)
 print("PROCESS_MEDIA_ENQUEUE_LOCALLY_IMMEDIATELY", PROCESS_MEDIA_ENQUEUE_LOCALLY_IMMEDIATELY)
 
+print("DIFFGRAM_MINIO_ENDPOINT_URL", DIFFGRAM_MINIO_ENDPOINT_URL)
```
PJEstrada, line 184: I think we can remove these prints before merging.

PJEstrada (changes requested):
Hi @ppolxda 

Thanks a lot for this great contribution! This was a feature that several users have requested.

The code looks pretty good to me. I think the refactors for the URL generation methods helped a lot to keep the code cleaner as well as the connector inheritance.

Some final pointers before merging:

- There are some print statements that we can remove. (Added the comment)
- It would be great if you could share with use a small guide/tutorial for adding to our docs. Might be worth mentioning the time for expiration so that other users now that and adjust the expiry settings accordingly.

Other from that, thanks a lot for your contribution on this. I'm going to mark it as "changes requires" so that you can cleanup prints and add some basic docs and then we can proceed with merge.

Thanks!!

ppolxda:
> Hi @ppolxda
> 
> Thanks a lot for this great contribution! This was a feature that several users have requested.
> 
> The code looks pretty good to me. I think the refactors for the URL generation methods helped a lot to keep the code cleaner as well as the connector inheritance.
> 
> Some final pointers before merging:
> 
> * There are some print statements that we can remove. (Added the comment)
> * It would be great if you could share with use a small guide/tutorial for adding to our docs. Might be worth mentioning the time for expiration so that other users now that and adjust the expiry settings accordingly.
> 
> Other from that, thanks a lot for your contribution on this. I'm going to mark it as "changes requires" so that you can cleanup prints and add some basic docs and then we can proceed with merge.
> 
> Thanks!!

Hi, @PJEstrada, About using documents to inform users about expiry settings.

We can try to protect variables to solve this problem.

[https://github.com/diffgram/diffgram/pull/857/commits/891ec60da04c399556c88802bbf66bfebb0e1af0](https://github.com/diffgram/diffgram/pull/857/commits/891ec60da04c399556c88802bbf66bfebb0e1af0)

shared/settings/settings.py
```diff
@@ -57,6 +57,13 @@
 DIFFGRAM_AWS_ACCESS_KEY_SECRET = os.environ.get('DIFFGRAM_AWS_ACCESS_KEY_SECRET')
 IS_DIFFGRAM_S3_V4_SIGNATURE = env_adapter.bool(os.getenv('IS_DIFFGRAM_S3_V4_SIGNATURE', False))
 
+# Minio Settings
+DIFFGRAM_MINIO_ENDPOINT_URL = os.getenv('DIFFGRAM_MINIO_ENDPOINT_URL')
+DIFFGRAM_MINIO_ACCESS_KEY_ID = os.getenv('DIFFGRAM_MINIO_ACCESS_KEY_ID')
+DIFFGRAM_MINIO_ACCESS_KEY_SECRET = os.getenv('DIFFGRAM_MINIO_ACCESS_KEY_SECRET')
+DIFFGRAM_MINIO_DISABLED_SSL_VERIFY = os.getenv('DIFFGRAM_MINIO_DISABLED_SSL_VERIFY', None)
+DIFFGRAM_MINIO_DISABLED_SSL_VERIFY = False if DIFFGRAM_MINIO_DISABLED_SSL_VERIFY else None
```
PJEstrada, line 65: This condition seems to be returning `DIFFGRAM_MINIO_DISABLED_SSL_VERIFY` to False after setting it to `True`. Maybe line 66 can be removed and you can use `env_adapter.bool(` as in line 58

install_local_without_config.py
```diff
@@ -0,0 +1,174 @@
+# -*- coding: utf-8 -*-
```
PJEstrada, line 1: I like the idea of a quick install without config! However I think we can remove this as it can be out of scope for this PR.

Some suggestions for this though:
- Maybe we can reuse the `install.py` code to avoid maintaining 2 installation files. You can use a similar approach to the connectors with s3/minio.
- Another approach instead of a different script is just to set the default values to be the local filesystem that way users can do `install.py` and hit enter for all questions without thinking so much to get a full installation with basic configs.
- Even though local file system sounds like a good default, I still think it's important that we check the existence of MinIO on the users machine. Otherwise, new users might think they just need a raw filesystem.

PJEstrada (changes requested):
Hey @ppolxda

Thanks again for all the changes! I added a couple comments now. I think install.py idea is something we can move to a separate PR to keep this one just with the MinIO support scope.
So main changes before merging should be:
- Remove the new install.py to a new PR keeping in mind suggestions in comments.
- Double check a condition in `settings.py`
I think @anthony-sarkis  will also do a review based on the new signed url generation changes to approve this one.

anthony-sarkis:
I added some small updates 

### Enable Import browser
https://github.com/diffgram/diffgram/pull/857/commits/fea296ed14bc199a531043f3ed88ed6cfdf15e9a
image

### Enable Export browser
https://github.com/diffgram/diffgram/pull/857/commits/5c27d5f7bbd4b6b8aff56f2bcaf25f4d2072374d
minio export

I did a small refactor
https://github.com/diffgram/diffgram/pull/857/commits/762266e9dda034730418daf9d750172877d07ffc

on both the existing s3 one, and the new minio one.

@ppolxda 
Just to second what Pablo's saying, we really appreciate your contribution here, and so if the install aspect can be a seperate pull we can do that as what we call a "fast follow" meaning we intend to release it within a short time but it's still seperate to limit scope creep on each pull request.

### Export issue
[ ] One issue I did bump into, when testing export,
it seems like even when I send a request that uses the AWS connection
```
4 <methods.connectors.s3_connector.S3Connector object at 0x000002523492DA00> amazon_aws
```
It is somehow using the Minio!
Confirmed by powering off minio and getting:
```
exception details: "Could not connect to the endpoint URL: \"http://10.0.0.27:9000/diffgram/export/16_diffgram_annotations__source_directory_Default_datetime_2022-05-03T01%3A59%3A29.333431.json\""...
````
(Observed issue prior to refactor so that's unlikely to be cause)
I'm not sure exactly what the issue there is.

install_local_without_config.py
ppolxda: OK, commit has been removed.

anthony-sarkis (approved):
Approved but still subject to Pablo's approval

anthony-sarkis:
@ppolxda  Looks great overall to me. I will leave one small question on `install.py`  

I like your idea for the install with minimal config. In the future we can build that into a seperate pull request too.  Just as an FYI we are directionally thinking of building a UI process for install at some point, and there may also be other install things like "one click" on clouds, etc.  moved this to here to track it https://github.com/diffgram/diffgram/issues/862

@PJEstrada  Pending your approval

shared/settings/settings.py
anthony-sarkis: `env_adapter.bool()` is added

PJEstrada (approved):

PR #887, opened 2022-06-02
Add schema migration (e.g. for labelbox)

anthony-sarkis:
1) Adds Label schema support to migrator
2) Improves security around schema and connection
3) Covers an edge case 
https://github.com/diffgram/diffgram/commit/5c3965994c04103f26199a4cb0a3ba4884a946d9
4) Improves use of regular methods to avoid session.commit()
5) Makes new object (label, attribute) more generic for schema https://github.com/diffgram/diffgram/commit/dfc7f74d463b416b914e1d9c2bea1d02080c09ac
So a new dev does not have to know about schema concept besides that it's a required value in function
6) Other little things, like swapping list auto complete, better detection for connection exists  etc

In the current iteration the user selects the schema from the UI instead of automatically being created from say labelbox's thing. My rationale is:
* It was faster to make then trying to make assumptions about automatically creating it.
* In a V2 or future iteration we can add "auto detect schema" as another feature
* Some users may want that control
* Schema concepts are more complex then dataset concepts, so it feels like something we should make less assumptions on
* If the user wants to manage it and keep schemas seperate they can, and if they want to just throw it all in one they can too

image


anthony-sarkis:
@PJEstrada I believe tests should be good now. I will await your review before merging

default/methods/project/project_new.py
```diff
@@ -78,14 +78,7 @@ def project_new_api():
             member_created=user.member
         )
 
-        schema = LabelSchema.new(
```
PJEstrada, line 81: Hey Anthony! I'm curious why the default schema creation was removed here?

Where is the label schema is created now when a new project is created?

shared/database/project.py
```diff
@@ -165,6 +165,14 @@ def new(session,
 
         session.flush()
 
+        schema = LabelSchema.new(
```
PJEstrada, line 168: Nevemind already found it :)

PJEstrada, line 168: Nevemind already found it :)

anthony-sarkis:
Thanks Pablo! 😊
Yes as you already found,
it's moved to be inside the new method so that a new engineer doesn't have
to think about the implications when they make the object.

PR #902, opened 2022-06-03
Keycloak Support For User Authentication

PJEstrada:
This PR adds a basic integration with keycloak to support federated user authentication.

For now, this is a completely optional integration and is not part of the diffgram core services. Although directionally we might want to add key cloak as part of our core installation services.
Docs:
https://dash.readme.com/project/diffgram/v1.0/docs/setting-keycloak-on-diffgram


anthony-sarkis:
Awesome!! Great work Pablo!! 😊 Have a great weekend see you Monday

PR #914, opened 2022-06-10
Ai workflow

anthony-sarkis:
#### Summary

- Refactor of connections to support new Workflow direction
- Work-in-progress of ML side

#### The Refactor of connections mostly includes
- `ConnectionStrategy` combines prior mix of files and manager class
- Connectors are moved to shared so they can be used by multiple services instead of just walrus
- Added `ConnectionStrategy.get_client()` method which is v1 of clarifying distinction between a concrete client like `text_analytics_client` and the diffgram `connection` database object.
- Added a `register(session)` to the actions base class. This calls `Action_Template.register()` which avoids a database migration for new actions. A new action can now simply add a `register(session)` and `register_all()` in `runners_mapping` will loop through them. I'm sure this will need further iteration but a step in hopefully right direction.

#### New pattern example
```
  connection_strategy = ConnectionStrategy(
      connection_class = AzureConnectorTextAnalytics,
      connector_id = connector_id,
      session = self.session)

  text_analytics_client = connection_strategy.get_client()
```
Now a connection class can be defined and not imported until required.

#### Work in progress items
Given the workflow overall is still a work in progress, these _may_ be safe to merge in current state
- adding scores per instance,
- starting points for actions, like a WIP of `AzureTextAnalyticsSentimentAction` and `HuggingFaceZeroShotAction`
- starting point (not fully wired) for manually trigging actions

*Edit
Currently assuming that `score` should not need to be added to the instance hash function because it's not changing.


anthony-sarkis:
@PJEstrada 
May I get your help please with the cypress test please,
There was only 1 failed, then merged master and 4.

For context this one https://dashboard.cypress.io/projects/djxpvm/runs/3575/overview/0edb3a01-6e07-47e3-b602-edf7416249a5  `diffgram/task_template/02_task_submit_to_review.cy.js`

I was able to reproduce the error by using the create sample data gen data
But at least for the UI case it fixed it here:
https://github.com/diffgram/diffgram/pull/914/commits/b19681709f120a39cf5f3abbe0e581bc233796f4

PJEstrada (approved):

Issue #899, opened 2022-06-03
Globals does not appear to expand on Schema change in studio


anthony-sarkis:
Workaround: Changing files appears to restore it
Priority: Low becuase appears only in studio, clear workaround, and new advanced feature

![globals in open state but not loading](https://user-images.githubusercontent.com/18080164/171922578-b8fcbc8f-ca5d-4142-a1c3-f579affa9a8d.PNG)

To reproduce:
1) Have two schemas.
2) One schema has globals, the other doesn't
3) Load a file with the non-global schema
4) Switch to schame with globals
5) Try to open globals tab

Appears to:
Not show globals

I thought this was a loading issue / sporadic but ran into enough times there is something there

anthony-sarkis:
This appears consistently in the Task context

anthony-sarkis:
For the attribute group type - but perhaps you already fixed it :)

PR #918, opened 2022-06-14
Fix global attributes on the Image interface

vitalii-bulyzhyn:
- [x] - #899


anthony-sarkis:
Thanks Vitalii, I tried this manually and now it appears to be working in tasks

I also discovered another bug during testing and believe it is fixed here https://github.com/diffgram/diffgram/pull/918/commits/0ba820d29076871acd73473f7f03593def4550e0

As soon as tests pass let's merge it

anthony-sarkis (approved):

PR #921, opened 2022-06-15
fix: improve dependencies of container in docker compose

PJEstrada:
Newer version of docker compose now allows to add extra conditions like checking if the service is healthy before starting the dependant containers.

This should help with the creation of the containers and avoid startup errors.

- Adds a new migration services that just runs alembic upgrade to ensure all migrations are runned correctly. Default service now wait till this is completed successfully.
- Upgrade docker compose to 3.9
- Adds better healthchecks for both postgres and rabbitmq to ensure both services are fully up before spinning up rest of containers.
- Remove env variables on the docker compose yaml and now only reference single .env, this improves the process of adding new env variables to the system since now its only necessary to add it on settings.py, install.py and helm chart (1 place less to place them)


anthony-sarkis (approved):
Excellent work Pablo!! 

I tested this manually and it worked really well, the rabbit and DB spin up, then when they are ready the migration runs, and then that smoothly exits.  Much better :)

PR #931, opened 2022-06-17
set to latest tag instead of specific version

anthony-sarkis:
Then we don't have to manually change the tag in order to upgrade
And it saves having multiple versions maybe in docker?
What do people think?

image


PJEstrada:
Hey Anthony! In general I think it could cause some issues to just set the `latest` tag on docker as any staging image that's built from a PR or some non-production builld would also override the latest tag. We can eventually think about something like our own tag called `newest` for example and just do custom logic on the publish docker actions to keep that one up to date.

anthony-sarkis:
Closing without merging as per discussion need more thought on this

PR #940, opened 2022-06-21
Fix Global Attributes Saving

PJEstrada:
Saving was not working properly due to wrong indexing during update of attributes.


anthony-sarkis:
Thanks @PJEstrada !!  

So was this case if there were globals and non-globals, and the global didn't load as 0th object?

Future: I wonder if we should think about moving away from indexes in this case and use creation ref / IDs instead.... maybe some small performance loss for extra look ups but avoids any risk of index mix ups

anthony-sarkis:
Thanks Pablo

PR #945, opened 2022-06-23
Improvements to OIDC Abstraction Layer

PJEstrada:
Add a new `OIDCProvider` Class to serve as an OIDC client factory. The idea is to be able to support different OIDC providers in the future across a standard interfaces and leave client-specific implementation details to the client specific libraries.


anthony-sarkis:
Excellent, this will give us a similar level of flexibility as the storage provider

PR #962, opened 2022-07-06
DAD-305 Add Connection Concepts to Ingestion

PJEstrada:
Allows the upload of data by specifying a custom connection ID. Files with the custom connection ID will be fetched directly from the connection source instead of being re uploaded into the default storage layer of diffgram installation.

This allows for blobs to be in multiple cloud providers at the same time, without duplicating storage on the diffgram storage layer.

This upload mechanism will only be supported via the SDK for now.


default/methods/attribute/attribute_template_group_update.py
```diff
@@ -23,7 +23,7 @@ def api_attribute_template_group_update(project_string_id):
         {'kind': str},
         {'label_file_list': None},
         {'default_value': None},
-        {'default_id': None},
+        {'default_id': Nlabel_file_listone},
```
anthony-sarkis, line 26: merge conflict `None`

shared/database/source_control/file.py
```diff
@@ -372,44 +377,59 @@ def get_geo_assets(self, session) -> list:
...
+            result.append(asset.serialize(session, connection_id = connection_id, bucket_name = bucket_name))
         return result
 
     def serialize_with_type(self,
                             session = None
                             ):
 
         file = self.serialize_base_file()
-
+        connection_id = self.connection_id
```
anthony-sarkis, line 392: Just a thought - if in the current context these values will always be self values,
then would it be better to have it as self. later in code?
This is good already. When I first read it though I assumed those values were being pulled from somewhere else. I'm wondering if it would be clearer to have ->

```
project = self.project,
connection_id = self.connection_id,
bucket_name = self.bucket_name)
```

shared/database/text_file.py
```diff
@@ -84,7 +87,6 @@ def serialize(self, session):
         }
         return text
 
-
     def regenerate_tokens_urls(self, session, new_offset_in_seconds):
```
anthony-sarkis, line 90: The tokens need to be upgraded to new pattern too I think

shared/url_generation.py
```diff
@@ -0,0 +1,172 @@
...
+    }
+
+    if connection.integration_name not in ALLOWED_CONNECTION_SIGNED_URL_PROVIDERS:
+        msg = f'Unsupported connection provider for URL regeneration {connection.id}:{connection.integration_name}'
+        log['error']['unsupported'] = msg
+        logger.error(msg)
+        return blob_object, log
+
+    connection_strategy = ConnectionStrategy(
+        connection_class = S3Connector,
```
anthony-sarkis, line 107: `connection_class` would be populated based on the connection itself right? maybe need like a look up table for that

walrus/methods/input/process_media.py
```diff
@@ -1646,38 +1662,36 @@ def process_one_image_file(self):
...
-        ### Main
-
-        self.resize_raw_image()
-
-        self.check_metadata_and_auto_correct_instances(
-            imageio_read_image = self.raw_numpy_image,
-            image_metadata = self.input.image_metadata
-        )
+        # Save thumbnails if we are uploading blobs, otherwise skip
+        if self.input.type != 'from_blob_path':
```
anthony-sarkis, line 1690: are we saving regening thumbnails in this context for future iteration?

walrus/methods/input/process_media.py
```diff
@@ -1646,38 +1662,36 @@ def process_one_image_file(self):
 
         """
 
-
-        result = self.read_raw_file()
-        if result is False: return False
+        # Read file if it does not come from a blob
+        if self.input.type != 'from_blob_path':
```
anthony-sarkis, line 1666: let's use the strategy pattern here too 
`strategy = 'default'` or `strategy = 'connection`' feels easier to reason about.

Pablo I'm guessing you would agree but more for others: here the read_raw_file() it's fairly obvious maybe not much difference, but lower down  `shutil.rmtree(self.input.temp_dir)` or not based on the path requires understanding a lot more context

shared/url_generation.py
```diff
@@ -0,0 +1,192 @@
...
+    :return:
+    """
+    if not blob_object.url_signed_blob_path:
+        return
+    should_regenerate, new_offset_in_seconds = data_tools.determine_if_should_regenerate_url(blob_object, session)
+
+    if should_regenerate is not True: 
+        return
+
+    strategy = determine_url_regenerate_strategy(
```
anthony-sarkis, line 155: added example pattern https://github.com/diffgram/diffgram/commit/92268c223d9901dfa7432eb0c7104223fb1fba4b

anthony-sarkis (commented):
Great work Pablo!!

frontend/src/components/annotation/annotation_core.vue
```diff
@@ -518,17 +518,16 @@
...
+              <v-icon size="450">mdi-download-off</v-icon>
               <div v-if="file && file.image">
-                URL Attempted To be Used: {{file.image.url_signed}}
+               <p>
+                 URL Attempted To be Used: {{file.image.url_signed ? file.image.url_signed : "null"}}
+               </p>
               </div>
-
-               {{file_cant_be_accessed_error}}
+              <v_error_multiple :error="file_cant_be_accessed_error"> </v_error_multiple>
```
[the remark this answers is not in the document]
PJEstrada: Sure! Main reason for this is just aesthetics of the error, here we can better display the url and fill out the entire canvas width, height.

But if you think we should default to the `v_error_multiple` component we can refactor it too! In this case the error messages are from aws so they also don't have the format we expect on that component, but we can always parse it as appropriate.

frontend/src/components/connection/connection_form.vue
```diff
@@ -216,6 +216,31 @@
               ></v-checkbox>
             </v-col>
           </v-row>
+          <v-row
+            v-if="connection.integration_name && connection_form_specs[connection.integration_name].aws_v4_signature"
+            no-gutters>
+            <v-col>
+              <v-checkbox v-model="connection.aws_v4_signature"
```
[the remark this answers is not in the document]
PJEstrada: thanks Anthony!

shared/database/source_control/file.py
```diff
@@ -149,6 +149,13 @@ def video_parent_file(self, session):
     task_id = Column(Integer, ForeignKey('task.id'))
     task = relationship("Task", foreign_keys = [task_id])
 
+    # Connection ID where the file is stored (if no connection provided will use default storage provider from env vars)
+    connection_id = Column(Integer, ForeignKey('connection_base.id'))
+    connection = relationship("Connection", foreign_keys = [connection_id])
+    bucket_name = Column(String())
+
+    storage_strategy = Column(String(), default="global_env_vars")   # ['connection', 'global_env_vars']
```
anthony-sarkis, line 157: @PJEstrada  I just added [this](https://github.com/diffgram/diffgram/commit/a30a4f0cf87b1d045b67d24e1d0fd105fea6d4f0) as an idea for your review... also not sure on naming I offered `default` in other case, but maybe it's better to define it directly like `global_env_vars`

walrus/methods/input/process_media.py
PJEstrada: Thanks Anthony! Agree that strategy is more readable. However are you picturing renaming the columns to this? Or are you picturing using this columns for the strategy and creating a completely separate function to ingest an image with conections?

walrus/methods/input/process_media.py
PJEstrada: For now we are not regenerating anything as we don't have the context if the connection will allow writing or if its desirable to write in the external bucket provided in the connection. So for now my  reasoning was to skip it. But we can further discuss altternatives you would like to try

anthony-sarkis: That makes sense! 
Future, I like the idea of (optionally) reading the blob still, and then writing the thumb artifact somewhere

walrus/methods/input/process_media.py
anthony-sarkis: Seperate functions would be better yes, then `process_one_image_file()` can call either strategy.

Skimming it looks like  moving `self.input.file = File.new(...)` to it's own `create_new_file()` function or similar would handle most of the shared stuff. I'm getting into the details here - up to you - but picturing something like

```
def process_one_image_file: 
  
    .... strategy routing ... then...

    save_connection_data()
OR
    save_default()...

def save_connection_data:
    create_new_file()
   ... connection specific stuff ...

def save_default:
    create_new_file()
   ... default specific stuff ...

```
** names are just for sake of example not suggesting these need to be actual names **

Not sure how much of the connection stuff there is specific to images, but may be some more cross over for other media types, may actually make some parts easier

frontend/src/components/annotation/annotation_core.vue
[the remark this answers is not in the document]
anthony-sarkis: OK all good - aesthetics is good, and we can parse better as future iteration here feels like a seperate issue then this pull request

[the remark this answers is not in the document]
PJEstrada: Ok sounds good! https://github.com/diffgram/diffgram/issues/984

shared/url_generation.py
PJEstrada: Super looks great will add this on the upload step too!

shared/database/source_control/file.py
PJEstrada: Added here: https://github.com/diffgram/diffgram/pull/962/commits/1525779d202783c6a7ee173bd760d8a48a79f380

shared/database/source_control/file.py
PJEstrada: Hey Anthony! I like more the first approach since we really enfoce we have all the data needed for actual url generation. In this case we might need have cases where storage_strategy might be one thing but does not have connection_name or bucket. I think the function gives us better guarantees for that if you're ok with it :)

PR #971, opened 2022-07-12
Cognito Token Storage Size & Login Optimizations

PJEstrada:
* Optimizes token storage size to be less than 4Kb in  order to keep using client side cookies. As a sidenote, if we ever want to move away from client side cookies we might need to think about server side storage of the session (on DB, redis, memcached or filesystem) or  localstorage storage of the session (not as safe as cookies)

* Change the logout flow to now have a button in login screen instead of directly login with SSO provider.


anthony-sarkis:
Good going Pablo! 😊

PR #974, opened 2022-07-13
Tree attribute optimisation 

vitalii-bulyzhyn:
Improve performance of tree search component


PJEstrada (approved):
Thanks Vitalii! The tree attribute is way faster now! I just tried it with 4K attributes and works like charm.

Notes so we can track them as separate issues for future:

- It would be cool if we could have fuzzy search so I can search things like (California 2015) and it shows up as a result.

vitalii-bulyzhyn:
@PJEstrada I also was thinking about that, can be good improvement for the future

PR #980, opened 2022-07-15
large api chunk size confiurable as env variable in frontend

ravi-dhime:
large file upload api is used by the frontend to upload a large file in chunks to walrus. walrus then uploads the same to data store and the subsequent processing takes place.

The chunk size is currently hardcoded to 5MiB in frontend. with minio, I noticed an issue with chunking. When such issues are seen, it would be good to be able to quickly change the chunk size for

quick debugging
trying out different chunk sizes as workaround
This PR proposes changes in frontend to read the chunk size from the container env variable.

Example usage in docker-compose.yaml:

```
   frontend:
    image: gcr.io/diffgram-open-core/frontend:${DIFFGRAM_VERSION_TAG}
+    environment:
+      VUE_APP_API_CHUNK_SIZE: 30
     ports:
     - "0.0.0.0:8081:80"
```

Tests:
tested locally in a docker-compose based setup using minio.


PJEstrada (changes requested):
Hey @ravi-dhime 

Thanks a lot for your contribution.

I wanted to suggest trying to access the variable in this way:


`console.log(process.env.VUE_APP_NOT_SECRET_CODE)`

I recall the `process.env` object has access to the env variables. That way we can keep the code on the actual index.js file and import it as usualy instead of writing a bash script that injects it to the file.

Can you please try this and let me know if it works for you?

ravi-dhime:
@PJEstrada, looks like the process.env gives the build time env variables and not the runtime env variables. I could not find any example that uses process.env for reading runtime env variables.

PJEstrada:
The more "standard" option we've used with other configs  is to set it on the backend `settings.py` and expose it via backend endpoint.

https://github.com/diffgram/diffgram/blob/master/default/methods/configs/is_open_source.py

Then consume it on frontend by adding a service:

https://github.com/diffgram/diffgram/blob/master/frontend/src/services/configService.js

ravi-dhime:
@PJEstrada , changed the approach to the standard option of diffgram.

configuration is via an env variable to the default service, which is exposed via a rest api. frontend service gets this value from the default service, once before every data upload. To update the chunk size value, need to change the env of default service and restart the service. No need to restart the frontend service, as it fetches the value before every upload.

PJEstrada (approved):
Great! This looks good to me thanks a lot @ravi-dhime

PR #981, opened 2022-07-15
hotkey for tag creation

ravi-dhime:
Tag can be created without touching the mouse. It is a good candidate to have a hotkey.


PJEstrada (approved):
Thanks good idea @ravi-dhime !

PR #996, opened 2022-07-21
Tag small improvements

anthony-sarkis:
-> Small improvements
-> Make job_list work in SDK context
Mainly https://github.com/diffgram/diffgram/commit/74189fdb7a9ca0fb0594c900891f94b0a097c27d

### DIFF
--- a/default/methods/task/task_template/job_list.py
+++ b/default/methods/task/task_template/job_list.py
@@ -18,9 +18,6 @@ from shared.database.tag.tag import JobTag
 
 @routes.route('/api/v1/job/list',
               methods = ['POST'])
-@General_permissions.grant_permission_for(
-    Roles = ['normal_user'],
-    apis_user_list = ["builder_or_trainer"])
 def job_list_api():
 
     spec_list = [{"metadata": dict}]
--- a/default/methods/task/task_template/job_list.py
+++ b/default/methods/task/task_template/job_list.py
@@ -88,8 +88,9 @@ def job_view_core(session,
 
     user = User.get(session)
 
-    if user.last_builder_or_trainer_mode != builder_or_trainer_mode:
-        raise Forbidden("Invalid user relation to builder_or_trainer_mode mode.")
+    if user:
+        if user.last_builder_or_trainer_mode != builder_or_trainer_mode:
+            raise Forbidden("Invalid user relation to builder_or_trainer_mode mode.")
 
     ### START FILTERS ###
 
--- a/default/methods/task/task_template/job_list.py
+++ b/default/methods/task/task_template/job_list.py
@@ -22,11 +22,6 @@ from shared.database.tag.tag import JobTag
     Roles = ['normal_user'],
     apis_user_list = ["builder_or_trainer"])
 def job_list_api():
-    # Would prefer to check all the inputs directly
-    # But then would have to revist whole metadata proposed concept
-    # for search
-    # TODO roll this into concept of reviewing regular method's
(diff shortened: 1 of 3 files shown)


default/methods/task/task_template/job_list.py
```diff
@@ -18,15 +18,7 @@
 
 @routes.route('/api/v1/job/list',
               methods = ['POST'])
-@General_permissions.grant_permission_for(
-    Roles = ['normal_user'],
```
PJEstrada, line 22: Do we want to make this endpoint public? Maybe we need to use the project permissions instead and modify the URL here?

default/methods/task/task_template/job_list.py
```diff
@@ -111,20 +86,9 @@ def job_view_core(session,
...
-        for tag in meta["tag_list"]:
-            if isinstance(tag, int):
-                tag_id_list.append(tag)
-        jobtag_list = JobTag.get_many(
-            session = session,
-            tag_id_list = tag_id_list,
-            project_id = project.id)
-        job_ids = [jobtag.job_id for jobtag in jobtag_list]
-        query = query.filter(Job.id.in_(job_ids))
+    query = add_tag_filters(query, meta, session, project)
```
PJEstrada, line 91: A lot cleaner now!

PJEstrada (commented):
Hey Anthony I did the review of the code! I'm just curious about why the permissions where completely removed here?

default/methods/task/task_template/job_list.py
anthony-sarkis: Thanks :)

default/methods/task/task_template/job_list.py
anthony-sarkis: Permissions are checked later by the project wrapper
The general permissions thing is only that the user exists, but it looks for a `user` object only not sdk auth.
This always relied on the project check, but to make it more clear I moved it higher up in execution order.

image
https://github.com/diffgram/diffgram/commit/74189fdb7a9ca0fb0594c900891f94b0a097c27d#diff-62217d126c56b801a3c7da723e459396e8a16950bab04f1502d1e507ebf0d5d5R67

I explained a bit about it here
https://github.com/diffgram/diffgram/commit/74189fdb7a9ca0fb0594c900891f94b0a097c27d

Another option reflecting on this now is we could move this to use project,
and then as we add other things in future that could have a separate endpoint, may be clearer permissions

anthony-sarkis: I had kind of wanted to just bump to the /project/project_string_id like we normally do but I also didn't want to conflict with upcoming auth changes.  Should we just move all the way to /project/project_string_id now?

PJEstrada: Ohh I see interesting! Thanks for pointing to the commit, I think we can leave it for now as it is and later we can discuss the "standard" way for permissions checking here. I think there might be cases where permissions can be checked before the actual endpoint handler, but there might also be cases were we'll need to extract more info to compute some other permissions.

hop-soellingeraj:
It looks like this failed QA in the merger.  Is that accurate?

PJEstrada:
> It looks like this failed QA in the merger. Is that accurate?

Our actions system tries to deploy to staging only if it finds a special keyword on the comments. Otherwise we 0/cancel the deployment so that's why its showing the red X. However all unit and integrations tests passed,so nothing to worry about :) https://github.com/diffgram/diffgram/actions/runs/2721166414

PR #1013, opened 2022-08-02
Fix project string

anthony-sarkis:
[bugfix](https://github.com/diffgram/diffgram/commit/3ebbce5f23b790c4cb91d8d44df478661c5d2547)
[Prevent confusing state](https://github.com/diffgram/diffgram/commit/c2b33ac4dde25b2dbd03c7c787dd3d1bdec804cb)


anthony-sarkis:
Small bugfix blocking for client merging becuase of that

PR #1036, opened 2022-08-24
Add attribute select element for the explorer

vitalii-bulyzhyn:
- Global attribute select
- Attribute select based on selected labels


PJEstrada (changes requested):
Hey Vitalii! 
Thanks for the code. Most of it looks good to me only comment:

for the name of attributes, whenever the name has whitespaces, please replace them with an underscore and that should make the query work

vitalii-bulyzhyn:
@PJEstrada thanks for pointing that

PJEstrada (approved):

PR #1042, opened 2022-08-31
Map data explorer

vitalii-bulyzhyn:
- add prototype of the map data explorer


vitalii-bulyzhyn:
Notes:
- need a proper way to display composable files (that are coming soon)
- need a proper way to display files that are on the same location (or close to each other), so points are not rendered for every single file

anthony-sarkis:
Given priority changes let's close this until re-focused / want to work on it

PR #1043, opened 2022-09-02
feat: jobs list display improvements and multiple select

PJEstrada:
* Add multiple select to status
* Add tags to table list
* Add schema name instead of labels list.


vitalii-bulyzhyn (approved):
Looks good to me

PR #1083, opened 2022-10-05
Core 2 job list UI improve pagination UI limit control

PedroIan:
I've changed the pagination to work on the server side instead of the frontend side. 

Considering the 2 ways of view (card view and list view), I had to add to methods of pagination on the frontend to work alongside with one pagination method on the server side.

Hope it fits Diffgram's needs and its clients' needs.


PJEstrada (changes requested):
Hey Pedro, thanks for the contribution:

Some notes after testing it:

1. There is no "loading" message when loading more elements on the card view. This makes it hard to know if we've reached the end of the job results or if there are more to be loaded.
2. We'll need to add unit tests and e2e tests to new backend and frontend changes.
3. The infinite scroller seems to not always work correctly. I try scrolling to bottom and it only works after 2 or 3 attempts (sometimes)
4. The table view does not work correctly with the Y-scrollbar, results get cut-off if I set it to 25-50 results or more per page.

I will merge this when have these features completely until then, I'll defer it until one of the team member has the time to work on it.

PR #1088, opened 2022-10-06
Fix us shema request

vitalii-bulyzhyn:
Fix creation of the new UI schema

### DIFF
--- a/frontend/src/components/ui_schema/ui_schema.ts
+++ b/frontend/src/components/ui_schema/ui_schema.ts
@@ -1,30 +1,32 @@
 import {v4 as uuidv4 } from 'uuid'
-
 export class UI_Schema {
-  public client_creation_ref_id
-  public client_created_time
-  public name
-  public archived
-  public is_visible
-  public is_public
+  public client_creation_ref_id: string
+  public client_created_time: string
+  public name: string
+  public archived: boolean
+  public is_visible: boolean
+  public is_public: boolean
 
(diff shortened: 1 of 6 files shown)


frontend/src/components/ui_schema/ui_schema.ts
```diff
@@ -1,30 +1,32 @@
...
     this.name = "Untitled";
     this.archived = false
     this.is_visible = true
     this.is_public = false
-
   }
 
-  public serialize(): any {
-    return JSON.stringify(this)
+  public export(): object {
```
PJEstrada, line 19: Instead of `object` here we can maybe define a type?

PJEstrada (approved):
Thanks a lot Vitalii!

frontend/src/components/ui_schema/ui_schema.ts
vitalii-bulyzhyn: But do we need it in this case? For now, that only place it used is in the request. Defining type will have sense if I do some refactoring like

**uischemaServices.js**
export const createUISchema = (ui_schema: UISchemaInterface) => {
       #code here
}

**ui_schema.ts**
public export(): UISchemaInterface {
# code here
}

**ui_schema_context_menu.vue**
await createUISchema(ui_schema.export())

What do you think? @PJEstrada

PJEstrada (approved):

PR #1091, opened 2022-10-07
Frontend improvements

vitalii-bulyzhyn:
Background on this PR (what we discussed earlier on Slack):

I've tried to think from the perspective of the new developer that may join Diffgram team or a developer who just want to contribute to Diffgram's code base, and there are few points that will confuse me (here will be example of tooltip button, but also applied for some other components):

- First thing I would see as a new developer is that **tooltip_button** has **ui_schema_name** property, and the first thing that I'll try to find out - what is that for (and most likely I will not get it without help of the team)
- Second, imagen that I contributes to Diffgram a new component that uses a **tooltip_button** and I wanna unit test it. I'm creating a test, run it - and it does not pass. The reason here will be that our component is coupled to **store** (but since I'm new to Diffgram, I don't know that), which will force me to mock a store - which will involve a lot more efforts to find out what exactly do I need to mock
- Third, as a new developer I'm coming to contribute to the one of the annotation interfaces, and I wanna add a new button to the **toolbar** I see implementations of schema ui management, and I'll ask how should I deal with this for my contribution

What I propose is to get rid of the handling schema ui inside the component and use a standardized  way of using **ui_schema** wrapper, which will allow us to better implement doc for the components with Storybook, write better tests, write better docs and make code more clear and more "contributable" for the new developers


anthony-sarkis:
Thanks Vitalii. I had forgot the UI Schema thing was manually wired in here. So that's right this does for sure need refactoring.

Some context here is "tooltip" button should really be renamed to something like "standard_button" since we roll various opinionated things into it, with the intent to create a standard button system wide.

If it's a wrapper or parameter either way the person would need to learn it at some point, so seeing it a structure in code or as a parameter doesn't really change that right?

If it's in the standard button, we could for example inherit the name of the button and pass that to UI schema. Basically inverse of de-coupling this, instead making it even more connected. The benefit is that a dev who "just wants a button" doesn't have to think about. In that case they wouldn't even need to know UI schema exists as a concept or param to add a new standard button.

Assuming that if we want UI schema, either way the vuex issue needs to be solved. So that veux coupling is for sure an issue, but an issue in either case.

So my main questions here are like: 
if we put the UI schema wrapper in the button component, would that achieve a similar thing?
how do we feel about the concept of a "standard button"? What direction do we see a concept like expanding into? 

What I really like about the proposal here is that the tool bar has various non-button elements, so for this specific instance it feels cleaner to have the wrapper external to the standard button. BUT not sure if that carries through to other parts of the application, or if it fits in long term goals of standard components for the most common functions

frontend/src/components/regular/ui_schema_wrapper.vue
```diff
@@ -1,39 +1,13 @@
...
+    <slot />
   </div>
 </template>
 
 <script lang="ts">
-
-/*
- 
- *
- *  EXAMPLE USAGE:
```
anthony-sarkis, line 20: why remove example usage? is it in docs now?

frontend/src/components/regular/tooltip_button.vue
```diff
@@ -109,148 +102,115 @@ export default Vue.extend( {
...
-    refresh_button_state_from_ui_schema(){
-      if (this.$props.ui_schema_name == undefined) { return true }
-      this.visible = this.$store.getters.get_ui_schema(this.$props.ui_schema_name, 'visible')
-    },
-    preventdefault(event) {
-      // we don't assume this will go anywhere unless opened in new tab
-      // in which case the event will follow the default. In the future
-      // could offer an "open in new tab" specific new
-
-      if (this.target != '_self'){ return } // eg '_blank' we want default opening in new tab
```
anthony-sarkis, line 236: 80% sure need this for opening in new tab (control click)
it's badly worded though, could maybe call it that directly and just do == _blank instead of the != which is confusing (I wrote it so just saying on myself)

frontend/src/components/regular/ui_schema_wrapper.vue
vitalii-bulyzhyn: The idea is to keep this on Storybook, will show you on the call

frontend/src/components/regular/tooltip_button.vue
vitalii-bulyzhyn: So, for now it was not used in the component, and I'd say if we don't use something maybe there is no sense to keep that for now?

vitalii-bulyzhyn:
Hey guys!

How I propose to start with Storybook is slowly start moving the most basic components to **base** directory, and make a rule that if the component is there it **MUST** be well documented in storybook

What do you think? @anthony-sarkis @PJEstrada

PJEstrada:
100% Agree Vitalii! Thanks for setting the first  ones up

frontend/src/components/regular/ui_schema_wrapper.vue
anthony-sarkis: Onboard for storybook!

frontend/src/components/regular/tooltip_button.vue
anthony-sarkis: L12 on prior it was used 
https://github.com/diffgram/diffgram/pull/1091/files/ce626834e2ad7c2808656a70b664a07bfd4f1422#diff-647d454dbc5ac5b6868117c257fe4f2173ee33d40ac018f273df99f567f1d69eL12

anthony-sarkis:
> Hey guys!
> 
> How I propose to start with Storybook is slowly start moving the most basic components to **base** directory, and make a rule that if the component is there it **MUST** be well documented in storybook
> 
> What do you think? @anthony-sarkis @PJEstrada

Agree.

The specifics to figure out will be where some of that documentation lives.
For examples it seems like here is some that can be done in each component/props like that description mentioned on slack ( https://github.com/diffgram/diffgram/pull/1091/commits/d173527a669b9b8d8145e67ccd5c185aae86c070) 
And then some of it can live in the examples we add for each thing in the `.stories`
AND there's a separate like doc page or doc function from storybook potentially.

So just to figuring out some patterns for that that fit our context may take a little bit of time but for sure worth it.

Thanks Vitalii for setting us on Playbook path

frontend/src/components/regular/tooltip_button.vue
vitalii-bulyzhyn: Sorry, my bad, lemme roll that back

PJEstrada (changes requested):
Thanks Vitalii! Just took a look at all the refactor and I think all is looking great!

The storybook  addition open a lot of new possibilities for the improvement of the frontend in general.

Only thing to check before merging is the unit tests that seem to be failing. Other from that looks good to me

PJEstrada (approved):
Hey Vitalii! I just checked out all the code and tested out the dir selector. Everything is working fine for me :)

Thanks a lot for all the effort on this one. All the code looks good to me so feel free to merge to master whenever ready.

PR #1092, opened 2022-10-07
CORE-22 Query Compound FIles on FIles Explorer

PJEstrada:
* Adds child files API endpoint to get all child files from a given parent ID file.
* Adds compound file preview component for query explorer
* Adds compound file preview on file list display for UI
* Modifies files list and query builder backend to include compound files 
* Added frontend types for Files, Image, Video, Audio, Geo and SensorFusion file types


anthony-sarkis:
Looks great Pablo!!

PR #1106, opened 2022-10-11
Fix: Uppercase Imports on GCP

PJEstrada:
* Centralizes supported file types
* FIxes bug on GCP imports with uppercase extensions


anthony-sarkis:
Awesome

PR #1131, opened 2022-10-20
CORE-46 Box Instance Refactor to V2 Types for Annotation Tools

PJEstrada:
## Description
This PR's objective is to refactor the entire Box instance under the new paradigms and type to be used for the improved version of Diffgram Annotation tools.

Main new Concepts:
- Interaction types.
- Coordinator Types
- Each Instance will now its own class and is only responsible for its inner state mutations.
- Record raw events and decorate them with additional annotation UI context information. `AnnotationToolEvent` type.

### Author Checklist
_All items are required. Please add a note to the item if the item is not applicable and please add links to any relevant follow up issues._

I have...

* [ ]  included the correct [type prefix](https://github.com/commitizen/conventional-commit-types/blob/v3.0.0/index.json) in the PR title
* [ ]  added `!` to the type prefix if API or client breaking change
* [ ]  provided a link to the relevant issue in JIRA or Github Link in PR.
* [ ]  included the necessary unit and integration
* [ ]  included comments & docs for new API Endpoints
* [ ]  updated the relevant documentation or specification in readme.io
* [ ]  reviewed "Files changed" and left comments if necessary
* [ ]  confirmed all CI checks have passed

### Reviewers Checklist
_All items are required. Please add a note if the item is not applicable and please add your handle next to the items reviewed if you only reviewed selected items._

I have...

* [ ]  confirmed the correct [type prefix](https://github.com/commitizen/conventional-commit-types/blob/v3.0.0/index.json) in the PR title
* [ ]  confirmed `!` in the type prefix if API or client breaking change
* [ ]  confirmed all author checklist items have been addressed
* [ ]  reviewed API design and naming
* [ ]  reviewed documentation is accurate
* [ ]  reviewed tests and test coverage
* [ ]  manually tested (if applicable)


frontend/src/components/vue_canvas/coordinators/coordinator_types/box/BoxInstanceCoordinator.ts
```diff
@@ -0,0 +1,28 @@
+import {Coordinator} from "../../Coordinator";
+import {BoxInstance} from "../../../instances/BoxInstance";
+
+export class BoxInstanceCoordinator extends Coordinator{
```
anthony-sarkis: What if this component's chief aim is to implement variations of  a `next(Interaction, Event)`function?

The main idea is to invert triggers based on mouse up/down. Instead to merge those as just a reason to "rerun" the loop. And the loop makes it's own choices based on Interaction + New Event.

So instead of processing a mouse down, it is about doing a 'query' that based on a given set of Interaction + New Event, what concrete functions (if any) should be called.

The test is to avoid having any type of JS event concepts as the primary path.

I'll make a concrete example to try and illustrate idea

`mouse_up()`
calls `ellipse_mouse_up()`

What if instead
1) The "global" router, is responsible from going from any type of interaction (just mouseup, mousedown to start), to say the appropriate type, here `ellipseCordinator`
(For example because Interaction.active_tool = 'ellipse' and ABC etc)

2) `ellipseCoordinator` will check a variety of conditions, to see if something should happen. For example, consider this existing function (slimmed to show point) 

```
ellipse_mouse_up():
      if (!this.ellipse_current_drawing_face && this.draw_mode) {
        this.$store.commit("init_draw");
      } else {
        this.$store.commit("finish_draw");
      }
```

We are already conditioning on if we are in the drawing state or not, to declare if we are starting or stopping drawing. 
So the fact that it's a mouse up, perhaps doesn't even matter? This function could be called equally well for mouse up or mouse down.

So for that function (ignoring use of vuex), if we just renamed it `ellipse.determine_drawing_state()` or something (just spitballing name), it could be called by either mouse up or mouse down.

I'm sure in practice there's some rough edges to that, but the point being that we should never rely on mouseup/mousedown as the primary assumption about the meaning of the interaction routing.

More generally
`def next()` ,
is based on set of `interaction.ABC` properties, to check if any coordinator actions are mapped to a set of those properties.
Which could include `interaction.mouse_state = something` or various properties.

The query could just be if statements, but the idea is the interaction itself is driving it, the mouse down is just one of the pieces of information it can use.

Let me know what you think :)

anthony-sarkis:
Great going Pablo!! 🎉
I really like the `AnnotationToolEvent` concept, the `Interaction` concept as it's own class, and treating each event as an addition to a log!!   I offered an idea on perhaps how we could think about the `Coordinator` class.

frontend/src/types/Interaction.ts
```diff
@@ -0,0 +1,29 @@
+import {AnnotationToolEvent} from "./AnnotationToolEvent";
+
+
+export class Interaction {
+  private event_log: AnnotationToolEvent[]
+  constructor() {
+    this.event_log = []
```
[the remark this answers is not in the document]
PJEstrada: Yes, in theory on each mouse_down, mouse_up, etc. We can append an `AnnotationToolEvent` object to the `Interaction` object. That way the unit tests can be mocked with specific events to simulate behaviours.

frontend/src/components/vue_canvas/coordinators/coordinator_types/box/BoxInstanceCoordinator.ts
PJEstrada: Thanks for sharing this example Anthony! I like the direction, one question though is where do you think this `next()` method is defined? I find the name a bit confusing since it feels like its trying to get the next event on the interaction but really the objective is to somehow consume the event and execute any actions based on the properties right? 

Right now I added the `process_mouse_down`, `process_mouse_up` and `process_mouse_move` to the Coordinator class. So the BoxCoordinator for example will process the mouse down like this:

```

  public process_mouse_down(): CoordinatorProcessResult {
    console.log('process_mouse_down BOX', this.box_instance)
    let result: CoordinatorProcessResult = {
      instance_moved: false
    }
    if(!this.box_instance && !this.annotation_event.is_actively_drawing){
      let currently_drawing_box: BoxInstance = this.annotation_event.current_drawing_instance as BoxInstance
      currently_drawing_box.set_actively_drawing(true)
      result.is_actively_drawing = true
    }

    return result;
  }
```

This is roughly similar to what you are proposing right? Just want to double check if this is the case or if maybe I'm missing something.

anthony-sarkis: Hey Pablo
On this coordinator class would be one option.
Some other naming ideas for it could be
* take_action_from_interaction
* map_interaction_to_functions
* get_next_concrete_function

The current implementation is starting with the js events, the mouse_up,down,move etc.
The proposal is to delete/remove those as the baseclass functions
and replace them with the above function, or something sort of in that direction

the "next" name may be bad, the theory was to signify that there is a series of events in the interaction log, and that it's taking the "next" step in the interaction, based on the new event. 

Let's discuss a bit on the call :)

frontend/src/components/vue_canvas/instances/BoxInstance.ts
```diff
@@ -0,0 +1,342 @@
...
+type BoxHoverPoints =
+  'x_min_y_min'
+  | 'x_max_y_min'
+  | 'x_min_y_max'
+  | 'x_max_y_max'
+  | 'not_intersecting_special_points'
+  | 'blocked'
+
+export class BoxInstance extends InstanceImage2D implements InstanceBehaviour2D {
+  public mouse_position: MousePosition;
```
vitalii-bulyzhyn, line 22: I'm a bit confused with this class. That seems like it also takes care of drawing and some other logic that is related to the annotation interface, and not to the instance itself, am I right?

frontend/src/components/vue_canvas/instances/BoxInstance.ts
```diff
@@ -0,0 +1,342 @@
...
+    this.initialized = true;
+    this.on_instance_hovered = this.set_default_hover_in_style
+    this.on_instance_unhovered = this.set_default_hover_out_style
+    this.ctx = ctx;
+  }
+
+  public duplicate_for_undo() {
+    let duplicate_instance = new BoxInstance(
+      this.mouse_position,
+      this.ctx,
```
vitalii-bulyzhyn, line 73: I though the idea would be stop using this concept of passing reference to canvas

vitalii-bulyzhyn (commented):
I'm a bit lost in the PR. What exactly `BoxInstance` should do? Why does it takes care of the drawing and we are still passing ref to canvas there?

frontend/src/components/vue_canvas/coordinators/InstanceTemplateCoordinatorGenerator.ts
```diff
@@ -29,35 +27,35 @@ export class InstanceTemplateCreationInteractionGenerator implements Interaction
...
-  private generate_mouseup_interaction(): Interaction{
+  private generate_mouseup_interaction(): Coordinator{
     let instance = this.get_hovered_instance();
     let current_instance = this.current_instance;
     if(!instance){
       instance = current_instance;
     }
     if(instance && instance.type === 'keypoints'){
-      return new KeypointInstanceMouseUp(instance)
+      return new KeypointInstanceCoordinator(instance as KeypointInstance, this.draw_mode)
```
vitalii-bulyzhyn, line 45: But why do we replace this?

PR #1227, opened 2022-11-29
Instance Store

vitalii-bulyzhyn:
# Instance Store

## Background

With the current architecture, every annotation interface keeps its own state and all the implementation of all the functionality. This was working fine a year ago, when Diffgram had only one annotation interface (Image + Video). Since then, we are growing rapidly, and now we have 5 (Image + Video, 3D, Text, Audio, Geo) different annotation interfaces, and potentially have more in the future, which will lead to more repeated code

Another reason, is a Compound File interface that are WIP right now. The problem here is that we should be able to annotate 2 different files simultaneously, with possibility to have same toolbar and same sidebar (and current architecture isn't ideal for this)

## Implementation

**InstanceStore** class is an management class that will store instances in an object with the key of file id:

```javascript
private _instance_store: any = {};
```

And there are two main methods: **get_instance_list** and **set_instance_list** (which are basically getter and setter). The example usage would be:

```javascript
this.instance_store.set_instance_list(this.file.id, newVal)
```

This will pass current instances to the management class, and they will be accessed on the parent component (**annotation_ui_factory.vue**):

```javascript
 this.instance_store.get_instance_list(this.file.id)
```

## File vs Task context

Since we have 2 different ways how file can be annotated (file and task), Instance Store should be able to work in the both modes. The Diffgram's main direction is move all the task related logic to the **annotation_ui_factory.vue**, and make annotation interfaces "context agnostic", however for now file id has been moved to computed property:
```javascript
    file_id: function() {
      if (!this.task) return this.file.id
      else return this.task.file.id
    },
```

### DIFF
--- a/frontend/src/components/annotation/annotation_core.vue
+++ b/frontend/src/components/annotation/annotation_core.vue
@@ -1094,7 +1094,7 @@ export default Vue.extend({
   },
   watch: {
     instance_list: function(newVal) {
-      this.instance_store.set_instance_list(this.file.id, newVal)
+      this.instance_store.set_instance_list(this.file_id, newVal)
     },
     finish_annotation_show: function (val) {
       if (val) this.annotation_show_on = false;
@@ -1586,6 +1586,10 @@ export default Vue.extend({
     };
   },
   computed: {
+    file_id: function() {
+      if (!this.task) return this.file.id
+      else return this.task.file.id
+    },
     label_file_map: function () {
       let result = {}
       for (let elm of this.label_list) {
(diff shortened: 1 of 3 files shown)


PJEstrada (approved):
Hey Vitalii! I really like the direction of this one. I think we can proceed to merge this one and start with the refactoring.

In terms of PRs I believe for the refactor we can keep the same approach of having small prs, but instead of merging to master we can merge to a different branch to make sure we do enough QA before merging to master.

PR #1234, opened 2022-12-01
Compound File base

vitalii-bulyzhyn:
## Description
...

### Author Checklist
_All items are required. Please add a note to the item if the item is not applicable and please add links to any relevant follow up issues._

I have...

* [ ]  added `!` to the type prefix if Breaking Changes.
* [ ]  considered the impact to the SDK.
* [ ]  included the correct [type prefix](https://github.com/commitizen/conventional-commit-types/blob/v3.0.0/index.json) in the PR title
* [ ]  provided a link to the relevant issue in JIRA or Github Link in PR.
* [ ]  included the unit and integration tests
* [ ]  included comments & docs for new API Endpoints
* [ ]  updated the relevant documentation or specification in readme.io
* [ ]  reviewed "Files changed" and left comments if necessary
* [ ]  confirmed all CI checks have passed

Breaking Changes including adding required params.

### Reviewers Checklist
_All items are required. Please add a note if the item is not applicable and please add your handle next to the items reviewed if you only reviewed selected items._

I have...

* [ ]  confirmed the Author checklist was followed
* [ ]  reviewed API design and naming
* [ ]  reviewed documentation is accurate
* [ ]  reviewed tests and test coverage
* [ ]  manually tested (if applicable)


frontend/src/components/annotation/annotation_core.vue
```diff
@@ -1098,26 +1167,17 @@ export default Vue.extend({
...
       deep: true,
       handler: function(newVal) {
-        if ((this.task && this.task.file && this.task.file.type === "video") || (this.file &&this.file.type === "video")) {
-          this.instance_store.set_instance_list(this.file_id, newVal)
-          if(this.task){
-            this.instance_store.set_file_type(this.file_id, this.task.file.type)
-          } else{
-            this.instance_store.set_file_type(this.file_id, this.file.type)
-          }
+        if (this.working_file.type === "video") {
```
PJEstrada, line 1178: Really cool! The `working_file` variable makes condition a lot more readable

PJEstrada (commented):
Hey Vitalii I just checked the code and I think it's going on a really good direction!

Eventually when v1 of this is released I think and important next step would be to take the `save`, `task_update`, `ghost_refresh_instances` , `ghost_determine_if_no_conflicts_with_existing`, `duplicate_instance_into_ghost_list`, and `build_keyframes_to_sequences_dict`, as well as a couple others into separate TS classes. This is just to avoid bloating now the UI factory class which is getting now to 1K lines.

I think we can easily create a couple classes based on the function you refactored. Some ideas are:


- `GhostInstancesManager`
- `VideoFramesController`
- `FileVideoSequences`

Just to name a few. I know right now might not be the ideal time for refactoring these as the main goal is just to refactor to a point where the compound file is possible, but when considering the rest of interfaces (text,3D, audio, etc). It's important that we don't default to move everything up to ui_factory but to try instead of creating a separate class for the functionality, otherwise we might end up with UI factory having logic for all the interface types which might end up bloating the component up a lot.

Anyways, those are my only comments, not really a thing for this PR but I think its important to keep it in mind.

Thanks a lot Vitalii overall work look really good! This is moving us closer to support the compound file.

frontend/src/components/annotation/annotation_core.vue
vitalii-bulyzhyn: Yeah, it's been set on the factory and now annotation_core can just do working_file.type for both task and file

vitalii-bulyzhyn:
Absolutely agree on the classes, I think we already discussed that on the call. But as you said, that will be v2 of the refactoring

anthony-sarkis:
Hey Vitalii, great work on big refactor!!
I sent review on Slack

PJEstrada (approved):
Thanks Vitalii! In terms of end user functionality most seems to be working! I added a couple extra ideas on slack but I think we can merge once the main bugs pointed by @anthony-sarkis  are fixed. I'll leave it approved and feel free to merge once the fixes are published

PR #1388, opened 2023-06-02
Box Performance Updates

anthony-sarkis:
For Labelbox Migration:

1) Batches Support
2) Fix project selection
3) Workaround for SignedURL access

Also updates to latest
and fixes UID from ID change in SDK

### DIFF
--- a/frontend/src/components/annotation/annotation_ui_factory.vue
+++ b/frontend/src/components/annotation/annotation_ui_factory.vue
@@ -1392,7 +1392,6 @@ export default Vue.extend({
                 video_data,
                 child_file_save_id: this.root_file.type === 'compound' ? this.annotation_ui_context.working_file.id : undefined
             }
-          console.log('save_request preeee')
             const [result, error] = await this.save_request(payload)
             if (result) {
                 // Save global instances video
--- a/frontend/src/components/annotation/annotation_ui_factory.vue
+++ b/frontend/src/components/annotation/annotation_ui_factory.vue
@@ -280,6 +280,7 @@
             </file_manager_sheet>
         </div>
         <v-dialog
+              v-if="dialog"
               v-model="dialog"
               max-width="290"
         >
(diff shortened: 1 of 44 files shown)


frontend/src/components/annotation/annotation_ui_factory.vue
```diff
@@ -1337,6 +1343,7 @@ export default Vue.extend({
                 )
             ) return
             this.set_save_loading(true, frame_number);
+            instance_list = to_serializable_instance_list(instance_list)
```
anthony-sarkis, line 1346: context:
circular dependency with new annotation classes
this removes that (object dependencies related to front end)

e.g errors when copying or saving

frontend/src/components/annotation/image_and_video_annotation/annotation_core.vue
```diff
@@ -3313,7 +3324,7 @@ export default Vue.extend({
 
       this.update_smooth_canvas(this.label_settings.smooth_canvas)
 
-      this.$forceUpdate();
+      await this.$forceUpdate();
```
anthony-sarkis, line 3320: makes initialization work correctly 

otherwise too heavy processing can have unexpected init problems in classes

e.g. canvas_mouse_tools being undefined (flaky) depending on time updating

frontend/src/components/vue_canvas/drawable_canvas.vue
```diff
@@ -218,7 +218,12 @@
         this.mouse_position,
         this.canvas_translate,
         this.canvas_element,
-        this.canvas_scale_global
+        this.canvas_scale_global,
+        this.instance_list,
+        this.canvas_height,
+        null,
```
anthony-sarkis, line 224: This is for previews
In this case not caring about zooms
Expectation is that it will then use full instance list

frontend/src/components/vue_canvas/instance_list.vue
```diff
@@ -269,11 +271,24 @@
...
           // MAIN loop
+          ctx.textAlign = "start";
+          ctx.setLineDash([0])
 
-          for (var i in this.instance_list) {
-            this.draw_single_instance(ctx, i)
+
+          // var startTime = performance.now();
+          if(!this.canvas_mouse_tools || this.video_mode){
+            for (let i in this.instance_list) {
```
anthony-sarkis, line 280: this is a "workaround" while we figure out best way to have the class structure with the visible thing

e.g. each instance has a `in_viewport` flag

frontend/src/components/vue_canvas/instances/BoxInstance.ts
```diff
@@ -275,7 +275,7 @@ export class BoxInstance extends InstanceImage2D implements InstanceBehaviour2D
     this.draw_circle(this.x_max, this.y_max, ctx)
     ctx.moveTo(this.x_min, this.y_max);
     this.draw_circle(this.x_min, this.y_max, ctx)
-    ctx.fill()
```
anthony-sarkis, line 278: loop calls `fill()` after anyway so this duplicate

PR #1536, opened 2023-08-22
New query route

francescov1:
## Description
- New query route to eventually replace `/file/list`. 
- Refactor File_Browser
- Added some unit tests and validation
- Small cleanups

### DIFF
--- a/default/methods/source_control/file/file_browser.py
+++ b/default/methods/source_control/file/file_browser.py
@@ -429,7 +429,7 @@ def view_file_list_web_route(project_string_id, username):
 
 @routes.route('/api/project/<string:project_string_id>/files', methods = ['GET'])
 @Project_permissions.user_has_project([
-    "admin", "Editor", "Viewer", "allow_if_project_is_public"])  # TODO: This array of perms seems commonly used. Can we create a constant for it somewhere to reduce boilerplate?
+    "admin", "Editor", "Viewer", "allow_if_project_is_public"])
 def api_get_files(project_string_id):
 
     with sessionMaker.session_scope() as session:
--- a/default/methods/source_control/file/file_browser.py
+++ b/default/methods/source_control/file/file_browser.py
@@ -714,7 +714,6 @@ class File_Browser():
                 return False
             file_count += count
         else:
-            # TODO: I think we can remove this now?
             policy_engine = PolicyEngine(session = self.session, project = self.project)
             perm_result = policy_engine.member_has_perm(
                 member = self.member,
--- a/default/methods/source_control/file/file_browser.py
+++ b/default/methods/source_control/file/file_browser.py
@@ -371,45 +371,6 @@ def get_instance_buffer(session, video_file_id, start, end):
 
     return buffer
 
-
-@routes.route('/api/project/<string:project_string_id>/files', methods = ['GET'])
-@Project_permissions.user_has_project([
-    "admin", "Editor", "Viewer", "allow_if_project_is_public"])
-def api_get_files(project_string_id):
-
-    with sessionMaker.session_scope() as session:
-
-        project = Project.get(session, project_string_id)
-
-        data = request.get_json(force = True)
-        query = request.args.get('query')
-
-        metadata_proposed = data.get('metadata')
-
-        if not metadata_proposed:
-            return jsonify("Error no metadata"), 400
-
-        # Add query to the proposed metadata so that File_Browser doesnt need to be changed
-        metadata_proposed['query'] = query
-
-        member = get_member(session)
-        
-        file_browser_instance = File_Browser(
-            session = session,
-            project = project,
-            metadata_proposed = metadata_proposed,
-            member = member
-        )
-        output_file_list = file_browser_instance.file_view_core(
-            mode = "serialize")
-
-        if len(file_browser_instance.log['error'].keys()) > 0:
-            return jsonify(log=file_browser_instance.log), 400
-
-        return jsonify(file_list = output_file_list,
-                       metadata = file_browser_instance.metadata), 200
-
-
 # Deprecated
 @routes.route('/api/project/<string:project_string_id>' +
               '/user/<string:username>/file/list',
@@ -453,6 +414,44 @@ def view_file_list_web_route(project_string_id, username):
             session = session,
             project = project,
             directory = directory,
+            metadata_proposed = metadata_proposed,
+            member = member
+        )
+        output_file_list = file_browser_instance.file_view_core(
+            mode = "serialize")
+
+        if len(file_browser_instance.log['error'].keys()) > 0:
+            return jsonify(log=file_browser_instance.log), 400
+
+        return jsonify(file_list = output_file_list,
+                       metadata = file_browser_instance.metadata), 200
+
+
+@routes.route('/api/project/<string:project_string_id>/files', methods = ['GET'])
+@Project_permissions.user_has_project([
+    "admin", "Editor", "Viewer", "allow_if_project_is_public"])  # TODO: This array of perms seems commonly used. Can we create a constant for it somewhere to reduce boilerplate?
+def api_get_files(project_string_id):
+
+    with sessionMaker.session_scope() as session:
+
+        project = Project.get(session, project_string_id)
+
+        data = request.get_json(force = True)
+        query = request.args.get('query')
+
+        metadata_proposed = data.get('metadata')
+
+        if not metadata_proposed:
+            return jsonify("Error no metadata"), 400
+
+        # Add query to the proposed metadata so that File_Browser doesnt need to be changed
+        metadata_proposed['query'] = query
+
+        member = get_member(session)
+        
+        file_browser_instance = File_Browser(
+            session = session,
+            project = project,
             metadata_proposed = metadata_proposed,
             member = member
         )
--- a/default/methods/source_control/file/file_browser.py
+++ b/default/methods/source_control/file/file_browser.py
@@ -372,6 +372,54 @@ def get_instance_buffer(session, video_file_id, start, end):
     return buffer
 
 
+@routes.route('/api/project/<string:project_string_id>/files', methods = ['GET'])
+@Project_permissions.user_has_project([
+    "admin", "Editor", "Viewer", "allow_if_project_is_public"])
+def get_files(project_string_id):
+
+    with sessionMaker.session_scope() as session:
+
+        project = Project.get(session, project_string_id)
+
+        data = request.get_json(force = True)
+        query = request.args.get('query')
+
+        metadata_proposed = data.get('metadata')
+
+        if not metadata_proposed:
+            return jsonify("Error no metadata"), 400
+
+        # Add query to the proposed metadata so that File_Browser doesnt need to be changed
+        metadata_proposed['query'] = query
+
+        user = User.get(session)
+        member = None
+        if user:
+            member = user.member
+        else:
+            # In some cases there can be no authorization (since project can be public)
+            # So we check if authorization exists in request.
+            if request.authorization:
+                client_id = request.authorization.get('username', None)
+                auth = Auth_api.get(session, client_id)
+                member = auth.member
+        file_browser_instance = File_Browser(
+            session = session,
+            project = project,
+            metadata_proposed = metadata_proposed,
+            member = member
+        )
+        output_file_list = file_browser_instance.file_view_core(
+            mode = "serialize")
+
+        if len(file_browser_instance.log['error'].keys()) > 0:
+            return jsonify(log=file_browser_instance.log), 400
+
+        return jsonify(file_list = output_file_list,
+                       metadata = file_browser_instance.metadata), 200
+
+
+# Deprecated
 @routes.route('/api/project/<string:project_string_id>' +
               '/user/<string:username>/file/list',
               methods = ['POST'])
@@ -426,7 +474,7 @@ def view_file_list_web_route(project_string_id, username):
         return jsonify(file_list = output_file_list,
                        metadata = file_browser_instance.metadata), 200
 
-
+# TODO: Would like to take the `directory` param out of this class, but there's a lot of dependencies on it still.
 class File_Browser():
     """
 
@@ -675,6 +723,7 @@ class File_Browser():
                 return False
             file_count += count
         else:
+            # TODO: I think we can remove this now?
             policy_engine = PolicyEngine(session = self.session, project = self.project)
             perm_result = policy_engine.member_has_perm(
                 member = self.member,
--- a/default/methods/source_control/file/file_browser.py
+++ b/default/methods/source_control/file/file_browser.py
@@ -375,7 +375,7 @@ def get_instance_buffer(session, video_file_id, start, end):
 @routes.route('/api/project/<string:project_string_id>/files', methods = ['GET'])
 @Project_permissions.user_has_project([
     "admin", "Editor", "Viewer", "allow_if_project_is_public"])
-def get_files(project_string_id):
+def api_get_files(project_string_id):
 
     with sessionMaker.session_scope() as session:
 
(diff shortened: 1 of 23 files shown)


default/methods/source_control/file/file_browser.py
```diff
@@ -372,6 +372,54 @@ def get_instance_buffer(session, video_file_id, start, end):
     return buffer
 
 
+@routes.route('/api/project/<string:project_string_id>/files', methods = ['GET'])
```
francescov1, line 375: This route is mostly a copy of the original route `/file/list` below, with a few tweaks (discussed in comments)

The route is now `GET /project/<project_id>/files` and accepts a query parameter called `query`. This feels more natural, since we are fetching a list of files and can optionally provide a query string to filter the files

default/methods/source_control/file/file_browser.py
```diff
@@ -372,6 +372,54 @@ def get_instance_buffer(session, video_file_id, start, end):
...
+        data = request.get_json(force = True)
+        query = request.args.get('query')
+
+        metadata_proposed = data.get('metadata')
+
+        if not metadata_proposed:
+            return jsonify("Error no metadata"), 400
+
+        # Add query to the proposed metadata so that File_Browser doesnt need to be changed
+        metadata_proposed['query'] = query
```
francescov1, line 393: To avoid changes to `File_Browser`, I simply added the query field to `metadata_proposed`. 

Wasnt sure how much of the `metadata_proposed` and `File_Browser` we want to change for the moment, so left them mostly intact for now

default/methods/source_control/file/file_browser.py
```diff
@@ -426,7 +474,7 @@ def view_file_list_web_route(project_string_id, username):
         return jsonify(file_list = output_file_list,
                        metadata = file_browser_instance.metadata), 200
 
-
+# TODO: Would like to take the `directory` param out of this class, but there's a lot of dependencies on it still.
```
francescov1, line 477: Doesnt make sense to pass a single directory to `File_Browser` since the queries support multiple directories. But there's a few areas in this class that depend on it, so left as is for now. Should I look at removing it?

default/methods/source_control/file/file_browser.py
```diff
@@ -675,6 +723,7 @@ def file_view_core(
                 return False
             file_count += count
         else:
+            # TODO: I think we can remove this now?
```
francescov1, line 726: I believe this policy check can be removed now, but not sure if anywhere else in code needs it

default/methods/source_control/file/file_cache_regen.py
```diff
@@ -1,12 +1,10 @@
 try:
     from methods.regular.regular_api import *
     from methods.source_control.file.remove import remove_core as file_remove_core
-    from methods.source_control.file.file_browser import File_Browser
```
francescov1, line 4: These imports were unused

shared/database/source_control/working_dir.py
```diff
@@ -170,6 +170,7 @@ def list(
         from shared.database.permissions.roles import ValidObjectTypes
 
         # Permissions: get datasets that user can see
+        # TODO: Can we remove this now?
```
francescov1, line 173: Another policy check that I think we can remove, but not sure if depended on elsewhere

shared/database/source_control/working_dir.py
```diff
@@ -278,13 +279,27 @@ def verify_directory_in_project(session, project, directory_id):
         Other ways we could do this but this seems to work as a basic check
         Future perhaps permissions on individual directories
         """
-        directory = session.query(WorkingDir).filter(
-            WorkingDir.project_id == project.id,
-            WorkingDir.id == directory_id).first()
-        if directory:
+        
+        directory = WorkingDir.get(session, directory_id, project.id)
+        return bool(directory)
```
francescov1, line 284: Small cleanup of the method above, since it was duplicate code from the `WorkingDir.get` method

shared/database/source_control/working_dir.py
```diff
@@ -278,13 +279,27 @@ def verify_directory_in_project(session, project, directory_id):
...
+    
+    # TODO: add tests
+    def can_member_view_datasets(session, project, member, dataset_ids):
```
francescov1, line 287: This is the new method to check if a member has access to a list of dataset IDs

shared/database/source_control/working_dir.py
```diff
@@ -278,13 +279,27 @@ def verify_directory_in_project(session, project, directory_id):
...
+        # TODO: How come we do this inline?
+        from shared.database.permissions.roles import ValidObjectTypes
```
francescov1, line 289: Curious why this is done inline vs top-level (copied from other Policy checks in this class)

shared/query_engine/expressions/dataset.py
```diff
@@ -18,6 +20,13 @@ def build_expression_subquery(self, session) -> Selectable:
         sql_column = query_op.column
         sql_compare_operator = self.operator.operator_value
 
+        can_view = WorkingDir.can_member_view_datasets(session = session, project = self.project, dataset_ids = raw_scalar_value, member = self.member)
+        
+        if not can_view:
+            # TODO: How to handle this better? Should we throw? 
+            # # Also seems like an awkard place to do validation in general, is there a better place? Difficulty is that we need access to the dataset ids to do this validation.
+            self.log['error']['unauthorized'] = f'You do not have access to these datasets'
```
francescov1, line 28: Not sure the best way to handle the error case here. Also feels awkward to add validation here, but we need access to the dataset IDs to do the validation in the first place

I've been a bit confused about error handling in general across the codebase. Seen a mix of raising errors, using loggers, etc. Would be good to do an overview call for error handling

shared/query_engine/expressions/expressions.py
```diff
@@ -46,6 +51,9 @@ def __init__(self,
         self.right_raw = right_raw
         self.compare_op_raw = compare_op_raw
 
+        self.member = member
+        self.project = project
```
francescov1, line 55: Added these two properties to the Expression class, I think there may be some more cleanup that can be done now that we have these available

shared/query_engine/query_creator.py
```diff
@@ -26,6 +26,7 @@ def __init__(self, session, project, member, directory = None):
         self.member = member
         self.directory = directory
         # Additional security check just for sanity
+        # TODO: This is duplicated from sqlalchemy_query_executor __init__, I think we can remove?
```
francescov1, line 29: I think this can be removed, exact same check is made in both the query creator and the query executor

shared/query_engine/diffgram_query.py
```diff
@@ -9,8 +9,7 @@ class DiffgramQuery:
         the DiffgramQueryExecutor.
     """
 
-    def __init__(self, tree, project, member, directory = None):
+    def __init__(self, tree, project, member):
```
francescov1, line 12: No need for directory property, and Im pretty sure it isnt used anywhere

default/methods/source_control/file/file_browser.py
francescov1: In general I wasnt sure how much we want to change `File_Browser` since its used in a few different places. We could just duplicate it for now so that we avoid breaking anything but start refactoring the new version of it?

default/tests/shared/database/source_control/test_working_dir.py
```diff
@@ -66,32 +68,102 @@ def test_list(self):
             'access_type': 'restricted'
         }, self.session)
 
+        mock_perm_result = Mock()
+        mock_perm_result.allow_all = False
+        mock_perm_result.allowed_object_id_list = [ds1.id, ds2.id, ds3.id]
+        mock_get_dataset_viewing_permissions.return_value = mock_perm_result
```
francescov1, line 74: `list` now depends on `get_dataset_viewing_permissions`, so we can just mock out that method rather than using real RoleMembers

default/tests/shared/database/source_control/test_working_dir.py
```diff
@@ -66,32 +68,102 @@ def test_list(self):
...
+
         dirs = WorkingDir.list(
             session = self.session,
             project_id = self.project.id,
             member = user.member,
         )
         dirs_id_list = [x.id for x in dirs]
-        self.assertEqual(len(dirs), 6)
+        self.assertEqual(len(dirs), 3)
```
francescov1, line 82: Not sure why this was checking `6` before, but with new mocking we only get the 3 directories created above

default/tests/shared/database/source_control/test_working_dir.py
```diff
@@ -66,32 +68,102 @@ def test_list(self):
...
-        self.assertEqual(len(dirs), 7)
+        self.assertEqual(len(dirs), 4)
         self.assertTrue(ds1.id in dirs_id_list)
         self.assertTrue(ds2.id in dirs_id_list)
         self.assertTrue(ds3.id in dirs_id_list)
         self.assertTrue(ds_restricted.id in dirs_id_list)
+
+    @patch("shared.database.source_control.working_dir.PolicyEngine")
+    def test_get_dataset_viewing_permissions(self, MockPolicyEngine):
+        # Arrange
```
francescov1, line 103: A pattern I commonly use is Arrange, Act, Assert. Arrange is for setup, mocking & defining variables, act is for triggering the action to test, and assert is for assertions.

The idea is to split every test method into these three sections. It helps ensure tests dont get too big, keeps them clean and makes it easy to understand for others. When you're expecting this pattern for a test, its much quicker to parse & understand it. 

Lemme know your thoughts!

default/tests/shared/query_engine/expressions/test_dataset.py
```diff
@@ -35,15 +35,46 @@ def setUp(self):
         self.auth_api = common_actions.create_project_auth(project = self.project, session = self.session)
         self.member = self.auth_api.member
 
-    def test_build_expression_subquery(self):
+    def initialize_compare_expression(self):
```
francescov1, line 38: Not sure if its common convention to define helper methods as part of the test class?

default/methods/source_control/file/file_browser.py
```diff
@@ -427,6 +427,43 @@ def view_file_list_web_route(project_string_id, username):
                        metadata = file_browser_instance.metadata), 200
 
 
+@routes.route('/api/project/<string:project_string_id>/files', methods = ['GET'])
+@Project_permissions.user_has_project([
```
francescov1, line 431: New route, only changes here are the new `query` URL param, and new route path `GET /api/project/<string:project_string_id>/files`

default/methods/source_control/file/file_browser.py
```diff
@@ -575,55 +612,58 @@ def __build_media_type_value(self) -> str:
...
-        limit_counter = 0
-        file_count = 0  # File count includes ones we don't actually query
-        # outside of limits...
-
-        if self.metadata['file_view_mode'] is None or \
-            self.metadata['file_view_mode'] not in ["changes", "annotation", "home", "task", "explorer", "base", 'ids_only']:
-            self.log['error']['file_view_mode'] = 'Invalid file_view_mode "{}"'.format(self.metadata['file_view_mode'])
-            return None
-        ignore_id_list = None
+    def get_query_file_list(self):
```
francescov1, line 615: I've pulled apart the `file_view_core` function into smaller functions, while trying to keep all logic the same. `file_view_core` still exists and does the same thing, but it calls a set of other functions instead of keeping all code inline.

I tried to leave all logic the same, and tested as much as I could, but someone who understands the full system should do a good QA to ensure I didnt break anything.

francescov1: The `get_query_file_list` method replaces [this block](https://github.com/diffgram/diffgram/blob/38d12c685c47a53061bcbe9b80c1ae257e651a84/default/methods/source_control/file/file_browser.py#L668-L675) of code

default/methods/source_control/file/file_browser.py
```diff
@@ -575,55 +612,58 @@ def __build_media_type_value(self) -> str:
...
+            limit = self.metadata["limit"],
+            offset = self.metadata["start_index"],
+        )
+        if not working_dir_file_list or regular_log.log_has_error(self.log):
+            return None, None
 
-        # For creating / viewing Jobs
+        return working_dir_file_list, count
+    
+    def get_job_file_list(self):
```
francescov1, line 625: `get_job_file_list` replaces [this block](https://github.com/diffgram/diffgram/blob/38d12c685c47a53061bcbe9b80c1ae257e651a84/default/methods/source_control/file/file_browser.py#L602-L625) of code

default/methods/source_control/file/file_browser.py
```diff
@@ -575,55 +612,58 @@ def __build_media_type_value(self) -> str:
...
-            # TODO future maybe just get ids only from sql
-            ignore_id_list = [i.id for i in file_list_attached_to_job]
+        return output_job_file_list, ignore_id_list
 
-            for index_file_attach, file in enumerate(file_list_attached_to_job):
-                file_serialized = file.serialize_with_type(session = self.session)
-                output_file_list.append(file_serialized)
-                output_file_list[index_file_attach]['attached_to_job'] = True
 
+    def get_all_file_list(self, ignore_id_list = None):
```
francescov1, line 655: `get_all_file_list` replaces [this block](https://github.com/diffgram/diffgram/blob/38d12c685c47a53061bcbe9b80c1ae257e651a84/default/methods/source_control/file/file_browser.py#L627-L665) & [this block](https://github.com/diffgram/diffgram/blob/38d12c685c47a53061bcbe9b80c1ae257e651a84/default/methods/source_control/file/file_browser.py#L678-L717) of code

default/methods/source_control/file/file_browser.py
```diff
@@ -575,55 +612,58 @@ def __build_media_type_value(self) -> str:
...
+        policy_engine = PolicyEngine(session = self.session, project = self.project)
+        perm_result = policy_engine.member_has_perm(
+            member = self.member,
+            object_id = self.directory.id,
+            object_type = ValidObjectTypes.dataset,
+            perm = DatasetPermissions.dataset_view
+        )
+        if not perm_result.allowed:
+            self.log['error']['unauthorized'] = f'Cannot view dataset ID: {self.directory.id}'
+            return None, None
```
francescov1, line 665: Previously, the following error handling would occur in this case:

```
resp = jsonify(log=self.log)
resp.status = 401
raise Unauthorized(response = resp)
```

now instead we just write to the log, since it allows us to pull out the error handling [here](https://github.com/diffgram/diffgram/pull/1536/files#diff-84ca682673d0f6d96b6c08959622d617aca63ab090e97754044e22cc6d5b717fR620). I tested the behaviour of both from the UI and found there was no difference, but let me know if its important to still raise an exception here

default/methods/source_control/file/file_browser.py
```diff
@@ -664,59 +704,33 @@ def file_view_core(
...
+            has_some_machine_made_instances = has_some_machine_made_instances,
+            ignore_id_list = ignore_id_list,
+            count_before_limit = True,
+            include_children_compound = self.metadata['with_children_files']
+        )
 
+        working_dir_file_list = query.all()
+        return working_dir_file_list, count
 
+    def add_pagination_metadata(self, file_count):
```
francescov1, line 733: `add_pagination_metadata` replaces [this block](https://github.com/diffgram/diffgram/blob/38d12c685c47a53061bcbe9b80c1ae257e651a84/default/methods/source_control/file/file_browser.py#L720-L729) of code

default/methods/source_control/file/file_browser.py
```diff
@@ -728,23 +742,68 @@ def file_view_core(
             else:
                 self.metadata['prev_page'] = None
 
-        if mode == "serialize":
-            for index_file, file in enumerate(working_dir_file_list):
-                if self.metadata['file_view_mode'] == 'explorer':
-                    file_serialized = file.serialize_with_annotations(self.session, regen_url = self.metadata["regen_url"])
+    def serialize_file_list(self, file_list):
```
francescov1, line 745: `serialize_file_list` replaces [this block](https://github.com/diffgram/diffgram/blob/38d12c685c47a53061bcbe9b80c1ae257e651a84/default/methods/source_control/file/file_browser.py#L732-L745) of code

default/methods/source_control/file/file_browser.py
```diff
@@ -664,59 +704,33 @@ def file_view_core(
...
+            ann_is_complete = ann_is_complete,
+            type = media_type_query,
+            return_mode = "query",
+            limit = self.metadata["limit"],
+            date_from = self.metadata["date_from"],
+            date_to = self.metadata["date_to"],
+            issues_filter = self.metadata["issues_filter"],
+            offset = self.metadata["start_index"],
+            original_filename = self.metadata['search_term'],
+            order_by_class_and_attribute = File.id,
```
francescov1, line 719: I noticed we have logic to determine `order_by_class_and_attribute` above, but then we dont use the result and just set the field to `File.id`. Is this a bug or on purpose?

PR #1538, opened 2023-08-28
fvirga/new query route

francescov1:
See https://github.com/diffgram/diffgram/pull/1536 for relevant comments & context (same diff, just created on a fork instead of direct on repo)

### DIFF
--- a/default/methods/account/plan/usage_cycle.py
+++ b/default/methods/account/plan/usage_cycle.py
@@ -260,7 +260,7 @@ class Usage():
 
         directory_list = WorkingDir.list(
             session = self.session,
-            project_id = self.project.id,
+            project = self.project,
             exclude_archived = self.exclude_removed
         )
         # Handle all directories in project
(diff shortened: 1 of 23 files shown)


default/methods/account/plan/usage_cycle.py
```diff
@@ -260,7 +260,7 @@ def determine_billable_directory_file_and_instance(
```
francescov1, line 260: See https://github.com/diffgram/diffgram/pull/1536 for comments explaining diffs

shared/query_engine/query_creator.py
```diff
@@ -25,13 +25,6 @@ def __init__(self, session, project, member, directory = None):
         self.session = session
         self.member = member
         self.directory = directory
-        # Additional security check just for sanity
-        Project_permissions.by_project_core(
```
anthony-sarkis, line 29: This is a duplicate. 
Context is that this is called later in SqlAlchemyQueryExecutor

Part of larger goal of aligning permissions into the same place

anthony-sarkis: Also that in general we are trying to put more validation at execution stage rather then prep/build stage (e.g dataset IDs which rely on the query string builder to format it first)

PR #1541, opened 2023-08-31
fvirga/new annotation unit tests

francescov1:
Added some new unit tests to Annotation_Update. Putting these on hold for now and instead working on the Annotation_Update refactor, so figure we can just merge the new tests in.

Ignore the huge diff for `default/tests/shared/test_annotation.py`, I just re-ordered the tests so that similar tests were grouped together

### DIFF
--- a/default/tests/shared/test_annotation.py
+++ b/default/tests/shared/test_annotation.py
@@ -125,6 +125,451 @@ class TestAnnotationUpdate(testing_setup.DiffgramBaseTestCase):
 
         self.assertEqual(ann_update.instance_list_kept_serialized[0]['sequence_id'], sequence.id)
 
+    def test_detect_and_remove_collisions(self):
+        label_file = data_mocking.create_file({'project_id': self.project.id, 'type': 'label'}, self.session)
+        file1 = data_mocking.create_file({'project_id': self.project.id, 'type': 'video'}, self.session)
+        frame = data_mocking.create_file(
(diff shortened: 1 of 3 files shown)


default/tests/shared/test_annotation.py
```diff
@@ -125,69 +125,69 @@ def test_update_sequence_id_in_cache_list(self):
```
francescov1, line 125: Nothing has actually changed in this file, I just re-ordered the tests so that test cases for the same methods are grouped together

default/tests/shared/test_annotation_new.py
```diff
@@ -0,0 +1,273 @@
...
+        # Act
+        instance = Annotation_Update(session=self.mock_session, file=self.mock_file, member=self.mock_member, project=self.mock_project)
+
+        # Assert
+        self.assertEqual(instance.log, mock_default_log.return_value) # Log was initialized
+        mock_Project_get_by_id.assert_not_called() # Project was not fetched
+        mock_get_member.assert_not_called() # Member was not fetched
+
+        # TODO: This isnt working, got it from here https://stackoverflow.com/questions/32463321/how-to-assert-method-call-order-with-python-mock
+        # func_calls.assert_has_calls([call.m1(), call.m2(), call.m3(), call.m4(), call.m5(), call.m6()])
```
francescov1, line 42: Wanted to add an assert to ensure we call each of these functions in the correct order, but couldnt get this strategy to work. Will revisit later if we still depend on the ordering.

PR #1561, opened 2023-10-19
Fix hotkeys for Multi images (Compound files)

sergzak022:
## Description
Fix a bug with hotkeys that happens in compound file context with multiple images. Updated HotkeyListener documentation to make it easier to use.

Bug Description:
Pressing a hotkey makes registered hotkey callbacks fire for all the image annotation panes

### DIFF
--- a/frontend/src/components/annotation/image_and_video_annotation/annotation_core.vue
+++ b/frontend/src/components/annotation/image_and_video_annotation/annotation_core.vue
@@ -741,6 +741,14 @@ export default Vue.extend({
   watch: {
     is_active: function (){
       this.canvas_element.style.cursor = ''
+      if ( !this.hotkeyListenerScope ) {
+        return
+      }
+      if ( this.is_active) {
+        this.hotkey_listener.setScopes([this.hotkeyListenerScope])
+      } else {
+        this.hotkey_listener.removeScope(this.hotkeyListenerScope)
+      }
     },
     global_instance: function(){
       this.$emit('global_instance_changed', this.working_file.id,  this.global_instance)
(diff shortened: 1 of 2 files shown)


frontend/src/components/annotation/image_and_video_annotation/annotation_core.vue
```diff
@@ -1733,6 +1741,10 @@ export default Vue.extend({
 
     this.setupHotkeys(this.hotkeyListenerScope)
 
+    if ( this.is_active ) {
```
sergzak022, line 1744: Once image annotation components are mounted, need to activate registered hotkeys ofr the active one.

frontend/src/components/annotation/image_and_video_annotation/annotation_core.vue
```diff
@@ -741,6 +741,14 @@ export default Vue.extend({
   watch: {
     is_active: function (){
       this.canvas_element.style.cursor = ''
+      if ( !this.hotkeyListenerScope ) {
+        return
+      }
+      if ( this.is_active) {
+        this.hotkey_listener.setScopes([this.hotkeyListenerScope])
+      } else {
```
sergzak022, line 749: The else block is needed for situations when user annotates in the compound context. For example text_annotation component currently uses a different mechanism for registering hotkeys. If user switches from image annotation to text in the same compound context, we needs to deactivate image annotation hotkeys.

sergzak022:
Hey Anthony, I'm not removing old hotkey manager class because it looks like it's still used by text_annotation_core component. Please see `listeners_map` method of `src/components/annotation/annotation_ui_factory.vue` and `keydown_event_listeners`/`keyup_event_listeners` methods of `text_annotation_core.vue`. Once we move `text_annotation_core` to use the new HotkeyListener, we can get rid of HotkeyManager.

anthony-sarkis:
Makes sense regarding keeping legacy `HotkeyManager` until text moves over to new one.

Tested manually and appears all working as expected now, thanks Sergey!

anthony-sarkis:
One small note for future is when we do remove legacy `HotkeyManager`, I believe we want to move the `is_active` scope setting to the higher level component (e.g. where `HotkeyManager` is currently sitting) to minimize code in "leaf" components.

anthony-sarkis (approved):

